PR #60, opened 2019-07-16
Add staging prefix to meadow bucket names

kdid:
You'll need to add `stack_with_environment` to your `staging.tfvars`

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -40,13 +40,13 @@ resource "random_string" "db_password" {
 }
 
 resource "aws_s3_bucket" "meadow_ingest" {
-  bucket = "${var.stack_name}-ingest"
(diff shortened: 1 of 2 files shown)


terraform/main.tf
```diff
@@ -40,13 +40,13 @@ resource "random_string" "db_password" {
 }
 
 resource "aws_s3_bucket" "meadow_ingest" {
-  bucket = "${var.stack_name}-ingest"
+  bucket = "${var.stack_with_environment}-ingest"
```
mbklein: Can we do something here like `"${var.stack_name}-${var.environment}-ingest"` so we don't have to worry about whether the stack part of stack_with_environment matches?

mbklein: (Then `environment` would just be `s` or `p`)

PR #66, opened 2019-07-17
Circleci deploy

mbklein:
Use CircleCI config & orbs to build and deploy meadow straight from CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,8 @@
 ---
-version: 2
+version: 2.1
+orbs:
+  aws-cli: circleci/aws-cli@0.1
+  docker-cache: nulib/docker-cache@0.1
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -57,8 +60,46 @@ jobs:
...
   ci:
     jobs:
       - test
+      - docker-cache/publish:
+          requires:
+            - test
+          filters:
+            branches:
+              only:
+                - /^deploy\/.+/
```
bmquinn: Will this eventually need to be `/^master|deploy\/.+/`?

mbklein: The branch list is an array, so it will eventually be:

```yaml
          filters:
            branches:
              only:
                - master
                - /^deploy\/.+/
```

bmquinn: Cool, thanks!

PR #78, opened 2019-07-19
Connect real inventory sheet list to Project page.  Stubbed out inven…

adamjarling:
…tory sheet display page.


kdid:
Does this restrict to the project? 

I was working on a route for that in this issue: https://github.com/nulib/repodev_planning_and_docs/issues/242 -  so I guess we can just update the route here when I'm done?

adamjarling:
@kdid Currently it's just pulling in ALL the inventory sheets.  I noticed the GET endpoint (specific to a Project) wasn't set up yet.  But should be easy enough to update the request once the Project specific route is in place.

PR #83, opened 2019-07-22
Create route for ingest jobs by project

kdid:
* change `ingest_job` routes to be nested under `project` resources


adamjarling:
React side looks good to me.

PR #84, opened 2019-07-22
Update ingest job filename key

kdid:
- change ingest job to store filename in this format: `s3://bucket-name/project-folder/filename.csv`
- remove `presigned_url` column from `ingest_jobs` table
- Update api call to only display ingest jobs for a particular project on Project screen

### DIFF
--- a/assets/js/components/InventorySheet/Form.jsx
+++ b/assets/js/components/InventorySheet/Form.jsx
@@ -35,16 +35,12 @@ const InventorySheetForm = ({ history, projectId }) => {
     try {
       const response = await axios.get("/api/v1/ingest_jobs/presigned_url");
       const presignedUrl = response.data.data.presigned_url;
-      const filename = presignedUrl
-        .split("?")[0]
-        .split("/")
-        .pop();
+      const filename = `s3://${presignedUrl.split("?")[0].split("/").slice(-3).join("/")}`
 
       await axios.put(presignedUrl, file, options);
       const ingestJobResponse = await axios.post(`/api/v1/projects/${projectId}/ingest_jobs`, {
         ingest_job: {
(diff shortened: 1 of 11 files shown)


assets/js/components/InventorySheet/Form.jsx
```diff
@@ -35,16 +35,12 @@ const InventorySheetForm = ({ history, projectId }) => {
     try {
       const response = await axios.get("/api/v1/ingest_jobs/presigned_url");
       const presignedUrl = response.data.data.presigned_url;
-      const filename = presignedUrl
-        .split("?")[0]
-        .split("/")
-        .pop();
+      const filename = `s3://${presignedUrl.split("?")[0].split("/").slice(-3).join("/")}`
```
mbklein: I'd really like to see us depend more on Elixir's [URI](https://hexdocs.pm/elixir/URI.html) module, especially since we're going to be needed to do a lot of file referencing. We also need to be prepared for the fact that Minio and AWS presigned URLs are going to look a little different. Minio's going to use `http://localhost:9001/bucket/path/to/key?signature` while AWS is going to use `https://bucket.s3.amazonaws.com/path/to/key?signature`. And our native S3 URI format will be `s3://bucket/path/to/key`.

So I guess the last 3 segments of the URI `path` will always be the key/filename, and maybe we don't need to parse the bucket if we're just using the configured one. But I still think it's a good idea to parse it as a URI and then take what we need instead of depending too much on `split()`.

kdid: I'm not following.

The line above is happening in React. So where do you want Elixir's URI module coming into play?

mbklein: Never mind. 😊

PR #121, opened 2019-08-14
Put query data into table rows, and start on Inventory sheet upload v…

adamjarling:
…alidation UI

This is a partial PR, but pushes the UI a little further down the road...

image

### DIFF
--- a/assets/css/meadow.css
+++ b/assets/css/meadow.css
@@ -34,16 +34,32 @@ a:hover {
   @apply underline;
 }
 
-input {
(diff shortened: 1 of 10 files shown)


adamjarling:
Yea, will do @kdid

kdid:
It probably would be helpful for us to store the original filename the more I think about it....

kdid:
The other thing I noticed is something weird going on with the title input box in the create project form. I swear it looked normal the first time I created a project but now it's "invisible". Checked both Safari and Chrome. If it's not happening for you, I'm ok with merging this and seeing if it's an issue on staging.

Screen Shot 2019-08-15 at 8 45 40 AM

PR #150, opened 2019-09-03
Support iex remote shell connections to running Meadow app

mbklein:
This PR contains the build/deploy-time changes and terraform code to allow IEx to attach directly to the running remote Meadow instance on staging for real-time introspection and debugging.


kdid:
Can we document instructions for how to connect?

PR #164, opened 2019-09-10
Require signin for Meadow

kdid:
- Using the app or API requires login with NU SSO:
  - Adds "Accounts" context in Meadow for user and groups related functionality
  - Adds AuthController  - SSO callback finds or creates user in database and stores some user information in session
  - SetCurrentUser plug intercepts all API requests and checks for user session. Puts the current user into the Absinthe execution context
  - Adds Absinthe Middleware `Authenticate` which checks for the current user in the context and if found resolves the query. 
  - Adds a GetCurrentUser query that returns the currently logged in user
  - Adds RequiresSignIn and CurrentUser components to the front end to guard the app

**Authorization was out of scope for this unit of work. For this PR anyone with a NetId can log in and use Meadow.


kdid:
@adamjarling - I added skip to one JS test as I wasn't sure how best to address.

assets/js/components/CurrentUser.js
```diff
@@ -0,0 +1,30 @@
...
+
+  return children(data.me);
+}
+
+CurrentUser.propTypes = {
+  children: PropTypes.func.isRequired
+};
+
+export default CurrentUser;
+export { GET_CURRENT_USER_QUERY };
```
adamjarling: What about creating a `/assets/js/components/Auth` directory, where we can organize Auth related components (`CurrentUser`, `RequireSignIn`) and queries?  Then we could create `.../Auth/auth.query.js` or something and avoid exporting from individual components?

kdid: Yeah. I think that's a good idea. Some of that stuff didn't seem to fit anywhere so I just threw it somewhere. Let me see if BMQ and MBK have comments and I'll change at same time.

lib/meadow_web/controllers/auth_controller.ex
```diff
@@ -1,19 +1,48 @@
...
+      {:error, _reason} ->
+        conn
+        |> redirect(external: referer)
+    end
+  end
 
-    body = OpenAM.uid(conn)
+  def logout(conn, _params) do
+    conn
+    |> delete_session(:current_user)
```
mbklein: For consistency's sake, can you use either `"current_user"` (as on line 24) or `:current_user` in both places? I prefer atoms wherever possible, but whichever.

Would atoms work for header names (`:content_type`, `:referer`) also?

kdid: the answer seems to be no on referer but I made the other change.

PR #172, opened 2019-09-13
added first pass of error state

davidschober:
added a rough of an error state on ingestion


adamjarling:
Looks good.  We can use this as a visual for how to map out the API.

image

PR #190, opened 2019-09-23
Add error messages to ingest job for front end to display

kdid:
- adds array of error messages for ingest sheet level errors 
- collapses "csv", "headers" and "file" states into "file" to for all top level errors
- Don't load ingest sheet rows into database unless the file level checks pass
- adds tests for different error messages
- removes csv extension check from ingest job (it's not working since the front end renames all files ".csv"). Printing the NimbleCSV error message instead.


adamjarling:
After discussing, we're going to change this "errors" key name to something a little more specific which points to file related errors...not row related.

image

Sounds like a `status` key will get added next for an Ingest Job, and there the `status` could be something to the effect of "Row error(s) in validation" or "File error(s) in validation".  Then the front end will look to that first for how to wire up component display logic.

image

PR #203, opened 2019-09-26
Super basic stubbed UI for Ingest Sheet Approved layout components

adamjarling:
Really basic.  I'd like to draw up some mock API data the front end could use here, and see if it meshes.

image

image

image


kdid:
This looks good, I'm thinking for terminology could we go with "Approved" and "Completed" (instead of Approved in progress and Approved) ?

PR #204, opened 2019-09-26
Move SQNS into its own top-level lib directory

mbklein:
* Move all SQNS/Pipeline components from `lib/meadow/ingest/pipeline` and `lib/meadow/utils/sqns` to `/lib/sqns`
* Move all associated tests and support files to reflect the new module structure
* Add support for custom attributes to be sent when kicking off an action pipeline
* Add [FilterPolicy](https://docs.aws.amazon.com/sns/latest/dg/sns-message-filtering.html) support to the SQNS configuration to allow conditional subscriptions based on custom attributes


mbklein:
The failing Coveralls test is due to moving a bunch of stuff around. If no one minds, let's merge and move on.

mbklein:
Please hold on this merge. One test file somehow got deleted and there are failures after putting it back.

mbklein:
OK, ready to go now.

PR #320, opened 2019-11-21
Implement sensible changes to test coverage

mbklein:
Addresses https://github.com/nulib/repodev_planning_and_docs/issues/306

- Add pipeline tests
- Remove JS from coveralls
- Relax test coverage options a little bit
- Use `mix excoveralls.html` locally for consistency

### DIFF
--- a/doc/architecture/decisions/0020-test-coverage-strategy.md
+++ b/doc/architecture/decisions/0020-test-coverage-strategy.md
@@ -1,0 +1,29 @@
+# 20. test-coverage-strategy
+
+Date: 2020-01-13
+
+## Status
+
+Accepted
+
+## Context
+
+Looking at our waning test coverage, we decided we needed a review of our test coverage
+strategy, especially a consideration of which parts of the project should be included
+in test coverage reports.
+
+## Decision
+
+* If low-level code is consistently coming up as uncovered, first check to
+  see if anything in the project is actually using that code. If not, remove
+  it.
+* Continue to write/run front end tests as needed, but remove them from the
+  coverage report.
+* Remove mocks, views, and GraphQL types from the coverage report.
+
+## Consequences
+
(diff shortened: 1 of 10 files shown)


doc/architecture/decisions/0020-test-coverage-strategy.md
```diff
@@ -0,0 +1,29 @@
...
+
+Looking at our waning test coverage, we decided we needed a review of our test coverage
+strategy, especially a consideration of which parts of the project should be included
+in test coverage reports.
+
+## Decision
+
+* If low-level code is consistently coming up as uncovered, first check to
+  see if anything in the project is actually using that code. If not, remove
+  it or at least comment it out.
```
kdid: I don't think I agree with leaving unused/untested code commented out. (Unless there is a very good reason in a specific case to do so.  We have a lot of extra code, a lot if it's boilerplate and would be easy to reimplement if we end up needing it. I'd say either delete it  -- or add tests and keep.

lib/meadow/ingest/rows.ex
```diff
@@ -8,35 +8,43 @@ defmodule Meadow.Ingest.Rows do
...
+  #
+  #      iex> update_ingest_sheet_row(row, %{field: bad_value})
+  #      {:error, %Ecto.Changeset{}}
+  #
+  #  """
+  #  def update_ingest_sheet_row(%Row{} = ingest_sheet_row, attrs) do
+  #    ingest_sheet_row
+  #    |> Row.changeset(attrs)
+  #    |> Repo.update()
+  #  end
```
kdid: See comment above. I'd vote for removing if not used.

doc/architecture/decisions/0020-test-coverage-strategy.md
mbklein: Updated to eliminate the option to comment it out.

PR #340, opened 2019-12-02
Eliminate redundant references to "Ingest..." throughout

mbklein:
This doesn't change any user-facing terminology – “Ingest Sheets” are still called “Ingest Sheets.” But the backend naming is _much_ cleaner, and some of it came over into the API as well.


mbklein:
Ah, right. You need to blow away your database with `mix do ecto.reset, meadow.pipeline.setup` before running this branch.

adamjarling:
Got it.  Everything seems to work on my end in the UI.  Nice job.

kdid:
So, I guess I thought this issue was more than just renaming `IngestSheets` to `Sheets` but also rethinking our whole approach to naming our Phoenix contexts and modules in general  - 
i.e. - what we decided here: https://github.com/nulib/meadow/blob/deploy/staging/doc/architecture/decisions/0015-phoenix-context-organization.md

Like example that `Meadow.Ingest.Projects` and `Meadow.Ingest.Projects.Project` was maybe complex and confusing 

Is that not the case?

mbklein:
That’s a good point. Yes, I think it’s worth revisiting (and refactoring) the names and depth of the contexts. I also think eliminating some of the dot-and-camel-case redundancy in the existing structure is a good start toward that. i.e., this PR doesn’t have to represent the whole of this refactoring work, but so far it looks like a good, easy cleanup that can start the process.

mbklein:
I just realized we missed a few renamings – directories and files, not modules and types, so it didn't mess up the tests – but please don't merge yet. 😄

kdid:
👍 Ok. Approving. Think we should leave issue #305 open though, since it doesn't really complete it.

PR #343, opened 2019-12-03
Mock out pyramid creation action

kdid:
- Creates shell of action for pyramid tiff generation
- Generates pyramid path pairtree using same format as donut
- Spawns mock node.js process passing source and destination
- Waits for response, mocks a completed status and a couple possible failures including a timeout if there is no response
- Adds pyramid bucket environment variable to config
- adds a node runtime image to dockerfile

### DIFF
--- a/lib/meadow/ingest/actions/create_pyramid_tiff.ex
+++ b/lib/meadow/ingest/actions/create_pyramid_tiff.ex
@@ -1,0 +1,76 @@
+defmodule Meadow.Ingest.Actions.CreatePyramidTiff do
+  @moduledoc "Create the pyramid tiff derivative for Image objects"
+
+  alias Meadow.Config
+  alias Meadow.Data.{AuditEntries, FileSets}
+  alias Meadow.Utils.Pairtree
+  alias Sequins.Pipeline.Action
+  use Action
+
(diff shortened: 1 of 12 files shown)


lib/meadow/ingest/actions/create_pyramid_tiff.ex
```diff
@@ -0,0 +1,76 @@
+defmodule Meadow.Ingest.Actions.CreatePyramidTiff do
+  @moduledoc "Create the pyramid tiff derivative for Image objects"
+
+  alias Meadow.Config
+  alias Meadow.Data.{AuditEntries, FileSets}
+  alias Meadow.Utils.Pairtree
+  alias Sequins.Pipeline.Action
+  use Action
+
+  @pyramid_bucket Config.pyramid_bucket()
```
mbklein: I don't think this module attribute ever gets used?

mbklein:
@kdid I made three small changes:
* added code to check for an existing open port before spawning the new one
* updated `AuditEntries.add_entry!` based on an interface change made in a previous merge
* rebased on staging (since @bmquinn merged #340)

Working great in my dev environment. Feel free to re-review if you want, but as far as I'm concerned, it's ready to merge.

PR #345, opened 2019-12-04
wip

kdid:
@adamjarling - this is where I've gotten so far. 

I've gotten Webpack to load the queries following the instructions here:
https://www.apollographql.com/docs/react/integrations/webpack/

But I haven't gotten it working with Jest, that same Apollo doc says to use this package:
https://github.com/remind101/jest-transform-graphql

I don't know yet whether it just doesn't work or we just don't have Jest all the way configured. 

(As an example, if it's not clear - I was just trying to implement the list all projects query/page)

### DIFF
--- a/assets/jest.config.js
+++ b/assets/jest.config.js
@@ -1,4 +1,8 @@
 module.exports = {
+  transform: {
+    "\\.(gql|graphql)$": "jest-transform-graphql",
+    ".*": "babel-jest"
+  },
   verbose: true,
   moduleNameMapper: {
     "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$":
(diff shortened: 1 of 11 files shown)


adamjarling:
Sweet!  Want me to pull it down and see if I can wrangle Jest into accepting `gql`?

kdid:
Sure!

adamjarling:
@kdid Making progress on this.  Have it working, but need to switch things up to support fragments now.  Will keep toying with it tomorrow.

adamjarling:
This seems helpful:  https://medium.com/@jialhe85/testing-tips-for-apolloclient-with-react-b58373eebb96

adamjarling:
So, the currently state of this branch is that for local testing, Jest now is able to process front end `.gql` files.  And webpack is able to process the files as well.   If we want to continue down this road, the next step would be to implement the `IntrospectionFragmentMatcher`:

https://www.apollographql.com/docs/react/data/fragments/#fragments-on-unions-and-interfaces

Using `.gql` fragments requires some additional set up.

It seems now is a good time to decide whether it's necessary or needed to store the `.gql` files on the front, and referenced by the back end.  After some discussion thought, this introduces coupling between the front and back ends, and might not be an independent pattern.  What if we have 4 UI apps hitting this GraphQL api...does the back end know about all of their local `gql` requests as well?

I'd probably lean to tabling this for now, and maybe re-evaluating after demo next week.

@kdid @mbklein @bmquinn

PR #346, opened 2019-12-04
Join ingest_sheet_rows to ingest_sheet_works instead of to itself

mbklein:
Fix Ecto join in order to retrieve the right number of ingest errors.


mbklein:
This may very well be the smallest PR I've ever submitted.

PR #358, opened 2019-12-10
Handle ingest sheet csv download on frontend

kdid:
- Downloads a CSV file of created works for use in metadata description 

This is considered temporary. We will probably need to do this from the backend at some point, but should be enough to put in front of users to gather feedback


README.md
```diff
@@ -34,6 +34,8 @@ If you need to reset the database you can run `mix ecto.reset` which will drop +
 
 If you just want to run the migrations but leave the data intact, you can just do `mix ecto.migrate`
 
+If you would like to use pgAdmin to view the database and tables, run `devstack up pgadmin` and view it in the browser at: [`http://localhost:5051/`](http://localhost:5051/)
+
```
adamjarling: Awesome.  Thanks for adding.

PR #359, opened 2019-12-10
Rename Audit Entries to Action States

mbklein:
* Renames all `AuditEntry`/`AuditEntries` modules to `ActionState`/`ActionStates`
* Renames all `audit_entry`/`audit_entries` files and directories to `action_state`/`action_states`
* Renames all `*_entry`/`*_entries` functions to `*_state`/`*_state`
* Renames Absinthe queries, types, fields, and resolvers as necessary
* Renames parameters and variables where it makes sense


lib/meadow_web/schema/types/ingest_types.ex
```diff
@@ -54,11 +54,11 @@ defmodule MeadowWeb.Schema.IngestTypes do
       resolve(&Resolvers.Ingest.ingest_sheet_rows/3)
     end
 
-    @desc "Retrieve all audit entries for an object"
-    field :audit_trail, list_of(:audit_entry) do
+    @desc "Retrieve all audit states for an object"
```
kdid: action?

kdid:
I don't think caused by this PR but when I try to query the action state using a work ID, I get this error:

```
   ** (UndefinedFunctionError) function CreateWork.actiondoc/0 is undefined (module CreateWork is not available)
        CreateWork.actiondoc()
        (meadow) lib/meadow_web/resolvers/ingest.ex:167: anonymous fn/1 in MeadowWeb.Resolvers.Ingest.get_action_states/3
```

mbklein:
> I don't think caused by this PR but when I try to query the action state using a work ID, I get this error:
> 
> ```
>    ** (UndefinedFunctionError) function CreateWork.actiondoc/0 is undefined (module CreateWork is not available)
>         CreateWork.actiondoc()
>         (meadow) lib/meadow_web/resolvers/ingest.ex:167: anonymous fn/1 in MeadowWeb.Resolvers.Ingest.get_action_states/3
> ```

You're right; it's unrelated to this PR but you found a bug. I'll fix it and test it in a separate PR.

PR #360, opened 2019-12-10
1129 rework groups

davidschober:
Added wireframes for groups, item, search, and batch edit. I kicked off a readme that is a start of documentation for how to contribute to wireframes.


davidschober:
creating a new pull request with cleaned up subjects because it was annoying me.

PR #385, opened 2019-12-17
Progress bar text is percentage complete

kdid:
- Make text on progress bar a percentage to match the progress bar itself.
- Git rid of artificial timeout on pyramid tiff generation (it will still mock the occasional error)

Screen Shot 2019-12-17 at 8 54 58 AM


adamjarling:
Seems cleaner on the implementation, and more appropriate for the generic component's purpose.

PR #390, opened 2019-12-18
Continue refactoring of Sheets module

kdid:
- Continued the work Brendan and I started on Friday refactoring working downwards in the `/lib/meadow/ingest/sheets.ex` file - moving things out that don't belong there. 
 - Created a Secondary context file for `status`. Moved (ingest) status related functions there. (not to be confused with the ingest sheet's status)
- Created a Secondary context file for `rows`. Moved rows related functions there. 
 - Created a Secondary context file for `sheet_works`. Moved sheet-works related functions to their own module - I went back and forth on this one because `sheet_works` is only a join table. (Not to be confused with sheetsToWorks which handles the works creation). However, it is one that we explicitly define and want the relationship data to remain separate from the ingest sheets themselves - so given that I decided it was appropriate to move some functions relating sheets to works and file sets there and it seems to help the size of the sheets context...
- Moved notifications related functions to a separate module.
- After talking to Adam attempted to disambiguate the two "progress" concepts we have related to ingest sheets (validation progress vs. ingest progress) by being explicit about naming validation progress related stuff. Also removed some code for validation subscriptions that isn't used as Adam/I agreed it is confusing. 

Didn't make it all the way down the file. There's still alot of action_states stuff that needs to move but PR getting too big.


assets/js/components/IngestSheet/IngestSheet.jsx
```diff
@@ -3,7 +3,7 @@ import { useQuery } from "@apollo/react-hooks";
 import Error from "../UI/Error";
 import Loading from "../UI/Loading";
 import IngestSheetValidations from "./Validations";
-import { GET_INGEST_SHEET_PROGRESS } from "./ingestSheet.query";
+import { GET_INGEST_SHEET_VALIDATION_PROGRESS } from "./ingestSheet.query";
```
adamjarling: Nice. Reads much better.

assets/js/components/IngestSheet/Report.jsx
```diff
@@ -6,8 +6,8 @@ import Loading from "../UI/Loading";
 import IngestSheetErrorsState from "./ErrorsState";
 import IngestSheetUnapprovedState from "./UnapprovedState";
 import {
-  GET_INGEST_SHEET_ERRORS,
-  GET_INGEST_SHEET_VALIDATIONS
+  GET_INGEST_SHEET_ROW_VALIDATION_ERRORS,
+  GET_INGEST_SHEET_ROW_VALIDATIONS
```
adamjarling: Again, much more clear to glance and quickly know whats going on.

PR #396, opened 2019-12-20
Add CRUD functionality to Collections

adamjarling:
This wires up Create, Update, and Delete for Collections.   Testing got kind of wonky as there's a lot of functionality packed into the CollectionForm component, so this is lacking in test coverage but working on some new patterns how to best handle testing against Apollo's MockProvider and routing changes using React Router's latest Hooks additions (ie. `useHistory()`, next.   I'll create a ticket for it as it will take some initial set up time.

Also note that the UI displays the latest wireframes for Collections data present in the UI, but the GraphQL schema only supports `id, description, keywords, name` values being mutated.

image

image

### DIFF
--- a/assets/css/meadow.css
+++ b/assets/css/meadow.css
@@ -157,6 +157,10 @@ table.clear th {
   @apply bg-white p-6 shadow-md;
 }
 
+.form-group {
+  @apply mb-4;
+}
+
 .icon {
   @apply fill-current inline-block text-gray-600 w-4 h-4 mr-1;
 }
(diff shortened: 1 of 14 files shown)


mbklein:
Create & Delete are working great. When editing, the `Collection Type` dropdown isn't getting pre-populated with the current value, so I can't Submit until I select one. (Safari 13.0.4 on OS X)

Meadow · Phoenix Framework 2019-12-23 12-27-04

adamjarling:
Yea that’s because the Collection Type isn’t yet supported on the back end.

adamjarling:
https://github.com/nulib/repodev_planning_and_docs/issues/398

PR #420, opened 2020-01-13
Allow work updating through API

kdid:
Adds `updateWork` mutation

### DIFF
--- a/assets/js/gql/FileSetFields.frag.gql
+++ b/assets/js/gql/FileSetFields.frag.gql
@@ -1,0 +1,11 @@
+fragment FileSetFields on FileSet {
+  id
+  accessionNumber
+  role
+  rank
+  position
+  metadata {
+    description
+    digests
+  }
+}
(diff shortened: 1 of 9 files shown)


assets/js/gql/FileSetFields.frag.gql
```diff
@@ -0,0 +1,11 @@
...
+  id
+  accessionNumber
+  role
+  rank
+  position
+  metadata {
+    description
+    digests
+  }
+}
```
adamjarling: Are we going to keep using these `gql` files?  The front end (for now) is not using them...just wondering if housing in the `/assets/` folder is best place?

kdid: We can move them. I'll open an issue to move them out of the front end.

mbklein: Maybe to `/test/gql/` or `/test/fixtures/gql/`?

kdid: @mbklein - can you add your thoughts to the issue?
https://github.com/nulib/repodev_planning_and_docs/issues/435

PR #451, opened 2020-01-27
Add ADR's for architecture decisions Elasticsearch indexing and IIIF …

kdid:


doc/architecture/decisions/0022-iiif-manifests.md
```diff
@@ -0,0 +1,22 @@
...
+
+Date: 2020-01-27
+
+## Status
+
+Accepted
+
+## Context
+
+Currently reqeusts for public manifests are routed through Donut in order to be re-cached in S3 if needed. We need to plan a strategy to handle IIIF manifest requests for manifests which were written by Meadow and should not be routed through Donut.
```
bmquinn:
```suggestion
Currently requests for public manifests are routed through Donut in order to be re-cached in S3 if needed. We need to plan a strategy to handle IIIF manifest requests for manifests which were written by Meadow and should not be routed through Donut.
```

PR #459, opened 2020-01-29
Add mock mutation for adding group to role

kdid:
@adamjarling - I forgot to mock the mutation.


mbklein:
> Great, sounds good. Looks like the build is failing but maybe that's because of something else?

Builds are fixed now. It was related to the Docker uid/gid issue in the LDAP container

PR #460, opened 2020-01-29
Complete LDAP.Entry schema

mbklein:
- Include a class and a subset of LDAP attributes in the `Meadow.Accounts.Ldap.Entry` struct
- Require full DNs for `list_group_members`, `list_user_groups`, `add_member`, and `remove_member`
- Use the new, improved `nulib/ldap-alpine` image in CircleCI

Resolves https://github.com/nulib/repodev_planning_and_docs/issues/471


test/support/ldap_case.ex
```diff
@@ -54,16 +55,39 @@ defmodule Meadow.LdapCase do
     :ok
   end
 
+  def display_names([]), do: []
+  def display_names(%Ldap.Entry{} = entry), do: entry.attributes.displayName
+  def display_names([entry | entries]), do: [display_names(entry) | display_names(entries)]
+
+  def ldn(val), do: "CN=#{val},OU=NotMeadow,DC=library,DC=northwestern,DC=edu"
+  def mdn(val), do: "CN=#{val},OU=Meadow,DC=library,DC=northwestern,DC=edu"
```
kdid: In the interest of clarity, do you think we could spell these out? `library_dn` and `meadow_dn`. (Only because when reading the tests I spent a couple minutes tracking this down before it occurred to me what the abbreviations meant.) Microsoft Developer Network?

mbklein: I had them spelled out initially, and changed them because I worried they made the tests look cluttered. I was very close to changing them back but left them. I’ll change them back.

PR #468, opened 2020-01-31
Use Exldap.Update for LDAP updates

mbklein:
Make it easier to write and maintain LDAP updating code by replacing `:eldap` calls with [`Exldap.Update`](https://hexdocs.pm/exldap_update/Exldap.Update.html#content)


mbklein:
Closing. Will re-submit later based on whatever happens with `Exldap.Update`.

PR #484, opened 2020-02-07
Bulma UI updates

adamjarling:
This will make the new UI work.  Definitely do a `yarn install` when testing locally.   It's about 90% of the way there, with more cleanup needed (which I'll get to), and also some functionality left less than perfect for Divya to jump in on.

When going over the code, you'll notice there are still some sections of previous code comment blocked out.  They're acting more or less as helpers as we finalize certain files, to remember what the original intent was.  They'll be gone in the next 2ish commits.


adamjarling:
Cool thanks @kdid. Without objection I’ll probably merge it this afternoon if that’s ok?

adamjarling:
I have 2 branches following it up...

PR #500, opened 2020-02-13
Prevent embedded nils

mbklein:
Make sure embedded nils are always converted to an empty map of the right type by the changeset


lib/meadow/data/schemas/validations.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Meadow.Data.Schemas.Validations do
+  @moduledoc """
+  This module provides custom changeset functions and
+  validations
+  """
+
+  @doc """
+  If a `cast_embed()` will result in a `nil` value (either on create or
+  update), set it to an empty embedded struct instead
+  """
```
kdid: Would this be a good module/function to try out a DocTest? I'd been looking for somewhere to do this - this seems like a good candidate.

PR #508, opened 2020-02-17
Add Work data to preservation tab

katdivyareddy10:
Added work data for preservation tab, modified Work query to extract fileset data for works.

This issue will be 100% complete when graphQL exposes Checksum, s3 key and verified values for the filesets.

### DIFF
--- a/assets/js/components/Work/Tabs/Preservation.jsx
+++ b/assets/js/components/Work/Tabs/Preservation.jsx
@@ -6,9 +6,7 @@ const WorkTabsPreservation = ({ work }) => {
   return (
     <div className="columns is-centered" data-testid="tab-preservation-content">
       <div className="column">
-        <p className="notification is-warning">
-          TODO: wire this table up to GraphQL data
-        </p>
+
         <table className="table is-fullwidth is-striped is-hoverable">
           <thead>
             <tr>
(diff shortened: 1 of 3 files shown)


assets/js/components/Work/Tabs/Preservation.jsx
```diff
@@ -21,20 +19,32 @@ const WorkTabsPreservation = ({ work }) => {
...
-                <FontAwesomeIcon icon="check" />
-              </td>
-              <td>
-                <button className="button">
-                  <FontAwesomeIcon icon="trash" />
-                </button>
-              </td>
-            </tr>
+
+            { work.fileSets && work.fileSets.length > 0 &&
```
adamjarling: No big deal, but just for the sake of practice, let's make one change to the PR code...   Here, the condition `work.fileSets.length > 0` technically isn't necessary since if the length === 0, any iterators on the array will just run 0 times.  How about you remove that and re-push?

adamjarling:
@katdivyareddy10 Can you also do the following?

- Connect the original issue to this PR
- Put a quick screen shot in the PR of what the work looks like
- Make a note of what work was not able to be completed, due to missing GraphQL data?   In addition, maybe you can search the board for an issue that looks like it represents exposing the missing fields in GraphQL, and if it doesn't look like anything matches, create a quick issue for adding the fields?

PR #511, opened 2020-02-18
Make sure the request method is an atom by the time it gets to the API request

mbklein:
Running in dev and test mode, `conn.method` always seems to be an atom (`:get`, `:post`, etc.). While investigating why the `/elasticsearch/` route was returning a 500 Server Error on staging, I discovered the following in the log:

```
2020-02-18 09:49:12 Request: GET /elasticsearch/meadow
2020-02-18 09:49:12 ** (exit) an exception was raised:
2020-02-18 09:49:12 ** (ArgumentError) argument error
2020-02-18 09:49:12 :erlang.atom_to_binary("GET", :utf8)
```

This PR coerces the method into an atom before sending it along to the Elasticsearch API

### DIFF
--- a/lib/meadow_web/plugs/elasticsearch.ex
+++ b/lib/meadow_web/plugs/elasticsearch.ex
@@ -21,7 +21,7 @@ defmodule MeadowWeb.Plugs.Elasticsearch do
       false ->
         conn
         |> put_resp_content_type("application/json")
-        |> resp(401, "Unauthorized")
+        |> resp(400, "Bad Request")
         |> send_resp()
     end
   end
--- a/lib/meadow_web/plugs/elasticsearch.ex
+++ b/lib/meadow_web/plugs/elasticsearch.ex
@@ -37,7 +37,13 @@ defmodule MeadowWeb.Plugs.Elasticsearch do
         Application.get_env(:meadow, Meadow.ElasticsearchCluster)
       )
 
-    case config |> config[:api].request(conn.method, path <> query_string, body, []) do
+    method = case conn.method do
+      x when is_atom(x) -> x
+      x when is_binary(x) -> String.to_atom(x)
+      x -> String.to_atom(to_string(x))
+    end
+
+    case config |> config[:api].request(method, path <> query_string, body, []) do
       {:ok, response} ->
         conn
         |> put_resp_content_type("application/json")
--- a/test/meadow_web/plugs/elasticsearch_test.exs
+++ b/test/meadow_web/plugs/elasticsearch_test.exs
@@ -11,7 +11,7 @@ defmodule ElasticsearchTest do
         |> auth_user(user_fixture())
         |> delete("/elasticsearch/meadow/_doc/#{work.id}")
 
-      assert conn.status == 401
+      assert conn.status == 400
     end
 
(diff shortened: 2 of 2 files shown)


lib/meadow_web/plugs/elasticsearch.ex
```diff
@@ -37,7 +37,13 @@ defmodule MeadowWeb.Plugs.Elasticsearch do
         Application.get_env(:meadow, Meadow.ElasticsearchCluster)
       )
 
-    case config |> config[:api].request(conn.method, path <> query_string, body, []) do
+    method = case conn.method do
```
kdid: Is the accepted method check above on line 17 still working?

lib/meadow_web/plugs/elasticsearch.ex
```diff
@@ -37,7 +37,13 @@ defmodule MeadowWeb.Plugs.Elasticsearch do
...
+      x when is_atom(x) -> x
+      x when is_binary(x) -> String.to_atom(x)
```
bmquinn: `String.to_atom` can get you into trouble with hitting the atom limit, I think it's a best practice to use `String.to_existing_atom` whenever possible instead (same goes for the next line).
```suggestion
      x when is_binary(x) -> String.to_existing_atom(x)
```

lib/meadow_web/plugs/elasticsearch.ex
mbklein: Good catch. Probably not. I’ll normalize it Up front when I get back.

mbklein: Good catch. Probably not. I’ll normalize it Up front when I get back.

lib/meadow_web/plugs/elasticsearch.ex
mbklein: `String.to_atom` will already reuse an existing atom if it’s been defined previously. The only difference is that `to_existing_atom` will throw an exception if the atom doesn’t already exist. I can’t see a situation in which that’s desirable in this case – we don’t have the option to proceed without an atom.

bmquinn: Right on, thanks for the clarification!

lib/meadow_web/plugs/elasticsearch.ex
mbklein: I’ve untangled what Plug, Elasticsearch, and HTTPoison are doing with respect to HTTP methods, and I think it’ll simplify things to go with them.
`Plug.Conn`’s method attribute is always an uppercase binary (e.g., `"GET"`) (per [Plug.Conn docs](https://hexdocs.pm/plug/Plug.Conn.html))
* Both Elasticsearch APIs (`HTTP` and `AWS`) pass the method unaltered to `HTTPoison.request/5`, which expects an atom. (per [HTTPoison docs](https://hexdocs.pm/httpoison/HTTPoison.html#request/5))

So we should be fine assuming the method is an uppercase binary on the way in (for checking against `@accepted_methods`) and atomizing before sending it along to `api.request/5`. The truly maddening part is that `HTTPoison` just wants to turn it back into a binary. 🙄

PR #513, opened 2020-02-19
Proposed Javascript refactor

kdid:
Proposed changes- 

- refactored the JS a bit for readability
- added a test for a failure case
- Moved the "s3://..." checking to the Elixir code.
  - It seemed like if the source url is wrong, don't spawn the process at all. 
  - It didn't seem necessary to check the target at all since we're generating it right there. But easy to add...

### DIFF
--- a/priv/tiff/pyramid.js
+++ b/priv/tiff/pyramid.js
@@ -1,99 +1,76 @@
 const AWS = require("aws-sdk");
-const Sharp = require("sharp");
+const sharp = require("sharp");
 const URI = require("uri-js");
 const fs = require("fs");
-const stream = require("stream");
 const tempy = require("tempy");
 
-function createPyramidTiff(source, dest) {
-  return new Promise((resolve, reject) => {
-    try {
-      makeInputFile(source)
-        .catch(err => reject(err))
-        .then(inFile => {
-          console.error(`Creating pyramidal TIFF from ${inFile}`);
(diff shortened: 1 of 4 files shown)


priv/tiff/pyramid.js
```diff
@@ -1,99 +1,70 @@
...
+      compression: "jpeg",
+      quality: 75,
+      tile: true,
+      tileHeight: 256,
+      tileWidth: 256,
+      pyramid: true
+    })
+    .toBuffer();
+  await sendToDestination(pyramidTiff, dest);
+};
```
adamjarling: Nice refactor.

kdid: Thanks @adamjarling - I just noticed I missed the `finally` clause that unlinks the file over to the left on line 39. Looking for where to stick that in...

mbklein:
Looks good. Way better than its nodejs 8.x ancestor. :smile:

I'd like to propose doing away with all the `console` logging – it's going to clutter our logs and our test suite without adding much. If we want to log status, maybe we can find a way to send it back through the port and let Elixir log it properly?

priv/tiff/pyramid.js
mbklein: Can you just wrap lines 9-26 in a `try { }` and then add a `finally { fs.unlink(inputFile) }` at the end?

PR #519, opened 2020-02-20
CRUD operations on collections are modified

katdivyareddy10:
- Modified Create and update pages to include all fields from collections.
- Added Delete option on the Screen when a particular collection is opened.
- Added Delete option on collection rows.
- Modified Read GraphQL statement.

Screen Shot 2020-02-20 at 10 45 34 AM

Screen Shot 2020-02-20 at 10 48 01 AM

### DIFF
--- a/assets/js/components/Collection/ListRow.jsx
+++ b/assets/js/components/Collection/ListRow.jsx
@@ -3,7 +3,7 @@ import PropTypes from "prop-types";
 import { Link } from "react-router-dom";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 
-const CollectionListRow = ({ collection }) => {
+const CollectionListRow = ({ collection, onOpenModal }) => {
   const { id, name = "", description = "", keywords = [] } = collection;
   return (
     <li data-testid="collection-list-row">
@@ -48,7 +48,7 @@ const CollectionListRow = ({ collection }) => {
               </Link>
             </p>
             <p className="control">
-              <button className="button" onClick={e => onOpenModal(e, project)}>
+              <button className="button" onClick={() => onOpenModal({id, name})}>
                 <FontAwesomeIcon icon="trash" />
               </button>
             </p>
--- a/assets/js/screens/Collection/List.jsx
+++ b/assets/js/screens/Collection/List.jsx
@@ -1,17 +1,24 @@
 import React, { useState, useEffect, useRef } from "react";
 import CollectionListRow from "../../components/Collection/ListRow";
 import { useQuery } from "@apollo/react-hooks";
-import { GET_COLLECTIONS } from "../../components/Collection/collection.query";
+import {
+  GET_COLLECTIONS,
+  DELETE_COLLECTION
+} from "../../components/Collection/collection.query";
 import Error from "../../components/UI/Error";
 import Loading from "../../components/UI/Loading";
-import { Link } from "react-router-dom";
-import debounce from "lodash.debounce";
+import { Link, useHistory } from "react-router-dom";
+import { useMutation } from "@apollo/react-hooks";
+import { useToasts } from "react-toast-notifications";
 import Layout from "../Layout";
+import UIModalDelete from "../../components/UI/Modal/Delete";
 
 const ScreensCollectionList = () => {
+  const { addToast } = useToasts();
   const { data, loading, error } = useQuery(GET_COLLECTIONS);
   const [filteredCollections, setFilteredCollections] = useState([]);
-  const [searchValue, setSearchValue] = useState("");
+  const [activeCollection, setActiveCollection] = useState();
+  const [modalOpen, setModalOpen] = useState(false);
   const inputEl = useRef(null);
 
   useEffect(() => {
@@ -20,12 +27,40 @@ const ScreensCollectionList = () => {
     }
   }, [data]);
 
+  const [deleteCollection] = useMutation(DELETE_COLLECTION, {
+    onCompleted({ deleteCollection }) {
+      addToast(`Collection ${deleteCollection.name} deleted successfully`, {
+        appearance: "success",
+        autoDismiss: true
+      });
+    },
+    refetchQueries(mutationResult) {
+      return [{ query: GET_COLLECTIONS }];
+    }
+  });
+
   if (loading) {
     return <Loading />;
   }
   if (error) {
     return <Error error={error} />;
   }
+
+  const onOpenModal = (collectionObj) => {
+    setActiveCollection(collectionObj);
+    setModalOpen(true);
+  };
+
+  const onCloseModal = () => {
+    setActiveCollection();
+    setModalOpen(false);
+  };
+
+  const handleDeleteClick = () => {
+    setModalOpen(false);
+    deleteCollection({ variables: { collectionId: activeCollection.id } });
+    setActiveCollection();
+  };
 
   const createCrumbs = () => {
     return [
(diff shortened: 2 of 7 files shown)


assets/js/screens/Collection/List.jsx
```diff
@@ -20,13 +30,41 @@ const ScreensCollectionList = () => {
     }
   }, [data]);
 
+  const [deleteCollection] = useMutation(DELETE_COLLECTION, {
+    onCompleted({ deleteCollection }) {
+      addToast(`Collection ${deleteCollection.name} deleted successfully`, {
+        appearance: "success",
+        autoDismiss: true
+      });
+      history.push("/collection/list");
```
adamjarling: Since we're already on this route in the application, we can remove the `history.push(...)` code to change the route.

adamjarling: And if not using elsewhere in the component, can also remove the `import` above.

assets/js/screens/Collection/List.jsx
```diff
@@ -1,17 +1,27 @@
...
+import UIModalDelete from "../../components/UI/Modal/Delete";
 
 const ScreensCollectionList = () => {
+  const { addToast } = useToasts();
   const { data, loading, error } = useQuery(GET_COLLECTIONS);
   const [filteredCollections, setFilteredCollections] = useState([]);
-  const [searchValue, setSearchValue] = useState("");
+  const history = useHistory();
+  const [deleteCollectionId, setDeleteCollectionId] = useState(null);
+  const [deleteCollectionName, setDeleteCollectionName] = useState("");
```
adamjarling: So this totally works fine, but I'd like to adjust it slightly if you're ok with it.  Instead of having separate `deleteCollectionId` and `deleteCollectionName ` state variables on the component, let's combine the two items into a single convenience object.    The object could be just passing back the entire `collection` object (if it looks reasonably light weight).  If the `collection` object looks like it contains an array of potentially a large number of rows, then probably easier to just define a simple object as argument to the callback function.  Something like?

`{ id: "COLLECTION_ID_HERE", name: "COLLECTION_NAME_HERE" }`

adamjarling: So what happens when the user hits "Cancel" on the Delete modal?   Looks like it retains the value of the "active" Collection.  Just to keep the state clean, let's clear out the value of the "active" Collection state variable on "Cancel".  You can just do something like `setActiveCollection(null)` or whatever.

assets/js/screens/Collection/List.jsx
```diff
@@ -1,17 +1,27 @@
...
 import Layout from "../Layout";
+import UIModalDelete from "../../components/UI/Modal/Delete";
 
 const ScreensCollectionList = () => {
+  const { addToast } = useToasts();
   const { data, loading, error } = useQuery(GET_COLLECTIONS);
   const [filteredCollections, setFilteredCollections] = useState([]);
-  const [searchValue, setSearchValue] = useState("");
+  const history = useHistory();
+  const [deleteCollectionId, setDeleteCollectionId] = useState(null);
```
adamjarling: No need to define a `null` value here.  Just use `useState()` to keep things clean.

adamjarling:
@katdivyareddy10 Also, please assign yourself to this PR (on the right under "Assignees").

assets/js/screens/Collection/List.jsx
katdivyareddy10: > All in all this looks really good. A few requested changes...
> 
> After you make and commit your changes, make sure to pull down the latest `deploy/staging` branch, then switch back to your feature branch `513-verify-collections-crud` and rebase on top of `deploy/staging`. The commands are:
> 
> ```
> git checkout deploy/staging
> git fetch
> git status (to see if you're up to date)
> git pull (if necessary)
> git checkout 513-verify-collections-crud
> git rebase deploy/staging
> ```
> 
> I'd like to get a situation eventually where there's a merge conflict, so we can walk through how to handle those. If you don't see any errors, then you can push up your changes.
> 
> Ideally we like to only include one commit per PR, so we can walk through how to squash your commits into one (if you made an extra commit for the requested changes).
> 
> Any questions, just ping me and we'll work through it.

Can I use this for combining commits ?
`git reset HEAD~
git add .
git commit -C ORIG_HEAD
git push -f origin`

assets/js/screens/Collection/List.jsx
katdivyareddy10: Ok, this is deleted! Can I re-assign this PR to you?

assets/js/screens/Collection/List.jsx
adamjarling: Yea for sure.  If going that route, go ahead and make your adjustments, then when it's all working as you like, run the above commands you mentioned.  When you force push your feature branch back up, `git push -f origin`, you may have to include your branch title as the final argument if you don't have your remote feature branch defined as an upstream branch.

assets/js/screens/Collection/List.jsx
```diff
@@ -20,13 +28,40 @@ const ScreensCollectionList = () => {
...
+  });
+
   if (loading) {
     return <Loading />;
   }
   if (error) {
     return <Error error={error} />;
   }
 
+  const onOpenModal = (collectionId, collectionName) => {
```
adamjarling: this parameter would then look like `const onOpenModal = collectionObj => {...`

assets/js/screens/Collection/List.jsx
```diff
@@ -20,13 +28,40 @@ const ScreensCollectionList = () => {
...
+    setActiveCollection({id: collectionId, name: collectionName});
```
adamjarling: And this becomes `setActiveCollection(collectionObj);`

assets/js/screens/Collection/List.jsx
```diff
@@ -20,13 +28,40 @@ const ScreensCollectionList = () => {
...
+    setModalOpen(true);
+  };
+
+  const onCloseModal = () => {
+    setActiveCollection({});
```
adamjarling: So, here's a weird (kinda) thing with JavaScript and empty objects.  `{ foo: "bar" } === true` and `{} === true`.  An empty object will evaluate to `true`.   Sometimes we do checks on variables like `if(someVar) { ... }`

So, let's change line 56 to be `setActiveCollection();` or if that doesn't seem to work, `setActiveCollection(null);`

katdivyareddy10: I'm just verifying something and realized the code fails if i setActiveCollection() to anything other than an empty object since i read this object ID on 
```
<UIModalDelete
        isOpen={modalOpen}
        handleClose={onCloseModal}
        handleConfirm={handleDeleteClick}
        thingToDeleteLabel={`Collection ${activeCollection.name}`}
      />
```

assets/js/screens/Collection/List.jsx
katdivyareddy10: This is a great idea, I'm a huge fan of one-liners! I'll do a re-check before i submit for PRs next time to see if I can think ES6 way!

assets/js/components/Collection/ListRow.jsx
```diff
@@ -48,7 +48,7 @@ const CollectionListRow = ({ collection }) => {
               </Link>
             </p>
             <p className="control">
-              <button className="button" onClick={e => onOpenModal(e, project)}>
+              <button className="button" onClick={() => onOpenModal(id, name)}>
```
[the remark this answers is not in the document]
katdivyareddy10: So, if i did onClick={openModal({id, name})} -- it goes into infinite loop and if I do onClick="openModal({id, name})" this obviously fails coz its getting a string, is there some other way to do this? Like use `` instead of "" or {}?

assets/js/screens/Collection/List.jsx
adamjarling: We discussed this offline and how to account for it with a ternary condition.

PR #520, opened 2020-02-20
Clean up display of Ingest Sheet completed state

adamjarling:
General cleanup of screens which feature a completed Ingest Sheet, but shows errors.  This PR also:

- Adds status to Ingest Sheet title on Ingest Sheet home screen
- Puts all Ingest Sheet errors in a nicely formatting, scrollable table
- Displays errors if they exist.  Previously a checkbox was used to show/hide errors, but these should be front and center to the user I'd guess?
- Adds in Northwestern Global marketing color palate.  
- Some new icons in the Navbar
- Updated footer

image

image

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 ## Initial setup:
 
 - Make sure you've done the [Local Authentication Setup](https://github.com/nulib/donut/wiki/Authentication-setup-for-dev-environment)
-- Install yarn if it's not already present: `brew install npm`, `npm -g install yarn`, or `asdf install yarn [VERSION]`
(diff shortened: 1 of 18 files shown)


adamjarling:
Thanks @kdid  I'd like to have @katdivyareddy10 have a look at the PR too.  If you had a few minutes, maybe you could walk her through how you check out PRs?   What to look for, whether to run locally or not, what to ignore or gloss over....best practices I guess?     This one is pretty huge, so maybe not the best example, but I think could serve the point a bit?

adamjarling:
@katdivyareddy10 Updated the styles with a quick fix.  I'd like to find a better solution when we have a bit more time..

image

PR #528, opened 2020-02-21
Configure Absinthe to use GraphQL Playground

kdid:
@adamjarling - @bmquinn mentioned switching from GraphiQL to GraphQL Playground a few weeks ago. It's a one liner configuration.

Could you and @katdivyareddy10 take a look and see if it seems better than GraphiQL to you? 

Screen Shot 2020-02-21 at 7 20 05 AM


adamjarling:
Looks good to me.  Let's give it a shot.

PR #536, opened 2020-02-24
Index sheet & project on work

mbklein:
- Create schema association between works & sheets
- Add `ingest` block to work index document
- Add insert/update/delete trigger on `ingest_sheet_works` association table to keep works in sync
- Don't forget to run `mix ecto.migrate` before using

### DIFF
--- a/.iex.exs
+++ b/.iex.exs
@@ -18,9 +18,10 @@ alias Meadow.Data.Schemas.{
 }
 
 alias Meadow.Ingest.{Sheets, Projects}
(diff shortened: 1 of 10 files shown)


mbklein:
Brendan found a [blog post](https://medium.com/flatiron-labs/til-how-to-select-merge-with-ecto-query-679d03204b9d) about using `select_merge` to associate certain fields with a schema without using a hard relation. The latest commit uses `select_merge` and a composable query to let `Meadow.Ingest.Sheets` inject the project and sheet names into a work struct without the work having to know about sheets or projects at all.

PR #542, opened 2020-02-25
Initial setup of Work screen tabs EDIT mode structure

adamjarling:
Partial PR for 410 which creates some initial setup for how we're wrapping a form around all tabs for Work Screen.   Once this is in place, it'd be a good time to pull in https://github.com/nulib/repodev_planning_and_docs/issues/427 as we start processing real form submissions.   After some patterns are set up there, I'll create sub-issues for @katdivyareddy10 to wire up each Work tab.


assets/js/components/Work/Tabs/About.jsx
```diff
@@ -1,59 +1,132 @@
...
+                  <p className="has-text-grey">
+                    {descriptiveMetadata.description ||
+                      "No description provided"}
+                  </p>
+                )}
+              </div>
+            </div>
+            <div className="field">
+              <label className="label">
+                Date Created (Should this be editable?)
```
kdid: Date created would not be editable.

I had kind of started an issue to determine what needs to happen to make work/file_set editing happen on the backend if you see anything now or later that needs to be added to that. 
https://github.com/nulib/repodev_planning_and_docs/issues/532

assets/js/components/Work/Tabs/Administrative.jsx
```diff
@@ -1,75 +1,99 @@
-import React from "react";
+import React, { useContext, useState } from "react";
 import PropTypes from "prop-types";
 import { Link } from "react-router-dom";
+import { WorkFormContext } from "../FormProvider";
+
+const visibilityOptions = ["RESTRICTED", "WHAT", "ARE", "THESE"];
```
kdid: These are an enum under the work type object. We could expose a way to get the list and terms through a query. For now you could probably hardcode the options.
Screen Shot 2020-02-26 at 9 06 45 AM

assets/js/components/Work/Tabs/Administrative.jsx
```diff
@@ -1,75 +1,99 @@
...
-            <option>With options</option>
-          </select>
+
+        <div className="field">
+          <label className="label">Preservation Level</label>
+          <div className="control">
+            {isEditing ? (
+              <div className="select">
+                <select>
+                  <option>Select dropdown</option>
```
kdid: These are the same as Donut: 1, 2, 3

https://github.com/nulib/donut/blob/master/config/authorities/preservation_level.yml

kdid:
Approving, adding some comments as to content of dropdowns, etc..

assets/js/components/Work/Tabs/Administrative.jsx
adamjarling: @kdid Got it.  I was kinda leaving those in for @katdivyareddy10 to find in Graphiql API.  I figured they'd be defined someplace, so good to know.

PR #544, opened 2020-02-25
Add specific titles to delete modal

katdivyareddy10:
Added specific titles for delete modal on Ingest Sheets.
Screen Shot 2020-02-25 at 1 49 25 PM

### DIFF
--- a/assets/js/components/IngestSheet/List.jsx
+++ b/assets/js/components/IngestSheet/List.jsx
@@ -54,7 +54,7 @@ const IngestSheetList = ({ project, subscribeToIngestSheetStatusChanges }) => {
   }
 
   const handleDeleteClick = () => {
-    deleteIngestSheet({ variables: { sheetId: activeModal } });
+    deleteIngestSheet({ variables: { sheetId: activeModal.id } });
     onCloseModal();
   };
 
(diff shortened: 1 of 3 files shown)


assets/js/components/IngestSheet/List.jsx
```diff
@@ -12,7 +12,7 @@ import { formatDate, TEMP_USER_FRIENDLY_STATUS } from "../../services/helpers";
 
 const IngestSheetList = ({ project, subscribeToIngestSheetStatusChanges }) => {
   const [modalOpen, setModalOpen] = useState(false);
-  const [activeModal, setActiveModal] = useState();
+  const [activeModal, setActiveModal] = useState({});
```
adamjarling: In about 99% of cases instead of initializing an empty object literal, it's better to just initial `false`, ie `useState()`.  You'll get into weird scenarios where you're testing against a false value when `{} === false; // is not true`.

assets/js/components/IngestSheet/List.jsx
```diff
@@ -64,7 +64,7 @@ const IngestSheetList = ({ project, subscribeToIngestSheetStatusChanges }) => {
   };
 
   const onCloseModal = () => {
-    setActiveModal(null);
+    setActiveModal({});
```
adamjarling: Same here as above comment.

assets/js/components/IngestSheet/List.jsx
katdivyareddy10: I edited it to reflect these changes.

PR #546, opened 2020-02-26
Added fileset images to Work page and Work page structure tab

katdivyareddy10:
I created a global function 'buildImageURL' to retrieve the correct image source and this function accepts the filesetId and imageSize as parameters. This would be useful to generate the source for any IIIF images in the future. Also added Tests for functions in Helpers.js

Screen Shot 2020-02-26 at 1 41 07 PM
Screen Shot 2020-02-26 at 1 41 21 PM
Screen Shot 2020-02-27 at 9 14 06 AM

### DIFF
--- a/assets/js/components/Work/ListItem.jsx
+++ b/assets/js/components/Work/ListItem.jsx
@@ -1,7 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { Link } from "react-router-dom";
-import { buildImageURL } from "../../services/global-vars";
+import { buildImageURL } from "../../services/helpers";
 
 const WorkListItem = ({ work }) => {
   const fileSetsToDisplay = 5;
@@ -16,6 +16,9 @@ const WorkListItem = ({ work }) => {
             <img
               src={`${buildImageURL(imageIdToDisplay, "IIIF_SQUARE")}`}
               alt="Placeholder image"
+              onError={e => {
+                e.target.src = "/images/1280x960.png";
+              }}
             />
           </Link>
         </figure>
--- a/assets/js/components/Work/ListItem.jsx
+++ b/assets/js/components/Work/ListItem.jsx
@@ -6,7 +6,7 @@ import { buildImageURL } from "../../services/helpers";
 const WorkListItem = ({ work }) => {
   const fileSetsToDisplay = 5;
   const imageIdToDisplay =
-    work.fileSets && work.fileSets[0] ? work.fileSets[0].id : "";
+    work.fileSets && work.fileSets.length > 0 ? work.fileSets[0].id : "";
 
   return (
     <div className="card">
--- a/assets/js/components/Work/ListItem.jsx
+++ b/assets/js/components/Work/ListItem.jsx
@@ -1,16 +1,22 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { Link } from "react-router-dom";
+import { buildImageURL } from "../../services/global-vars";
 
 const WorkListItem = ({ work }) => {
   const fileSetsToDisplay = 5;
+  const imageIdToDisplay =
+    work.fileSets && work.fileSets[0] ? work.fileSets[0].id : "";
 
   return (
     <div className="card">
       <div className="card-image">
         <figure className="image is-4by3">
           <Link to={`/work/${work.id}`}>
-            <img src="/images/1280x960.png" alt="Placeholder image" />
+            <img
+              src={`${buildImageURL(imageIdToDisplay, "IIIF_SQUARE")}`}
+              alt="Placeholder image"
+            />
           </Link>
         </figure>
       </div>
--- a/assets/js/services/helpers.js
+++ b/assets/js/services/helpers.js
@@ -9,10 +9,10 @@ import { IIIF_SIZES } from "./global-vars";
  * @return {string} This returns a URL for the image
  */
 export function buildImageURL(id, imageSize) {
-  const imageHost = id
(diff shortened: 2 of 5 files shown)


adamjarling:
This looks awesome.  We're adding some tests before merging the PR.

assets/js/components/Work/ListItem.jsx
```diff
@@ -1,16 +1,25 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { Link } from "react-router-dom";
+import { buildImageURL } from "../../services/helpers";
 
 const WorkListItem = ({ work }) => {
   const fileSetsToDisplay = 5;
+  const imageIdToDisplay =
+    work.fileSets && work.fileSets[0] ? work.fileSets[0].id : "";
```
adamjarling: Here's a little trick.  Usually when validating `array` type values, you'll check for length.  

`work.fileSets.length > 0 ? work.fileSets[0].id : ""`

assets/js/services/helpers.js
```diff
@@ -1,5 +1,19 @@
...
+import { IIIF_SIZES } from "./global-vars";
+
+/**
+ * Builds IIIF url for placeholder images
+ * @param {string} id This ID is a filesetId
+ * @param {object} imageSize This is an image size type
+ * @return {string} This returns a URL for the image
+ */
+export function buildImageURL(id, imageSize) {
+  const imageHost = id
```
adamjarling: How about we re-name this variable `imageHost` so it's a bit more semantic.  The variable value will be used to fill `<img src="VALUE HERE" ...` so why don't we name the variable something like `srcPath`?

assets/js/services/helpers.test.js
```diff
@@ -0,0 +1,75 @@
+import {
+  buildImageURL,
+  escapeDoubleQuotes,
+  formatDate,
+  getClassFromIngestSheetStatus
+} from "./helpers";
+
+describe("Build Image URL function", () => {
+  it("should return URL", () => {
```
adamjarling: Let's make this a little more specific.  How about something like "should return a URL containing the correct IIIF params..."

assets/js/services/helpers.test.js
```diff
@@ -0,0 +1,75 @@
...
+    const expected =
+      "http://localhost:8183/iiif/2/filesetId/square/500,500/0/default.jpg";
+    const actual = buildImageURL("filesetId", "IIIF_SQUARE");
```
adamjarling: This totally works fine, but if I was quickly scanning this code I may confuse "filesetId" for representing a variable value.  When passing around mock data in tests, make it something like the real data you're mocking, but more simple.   How about something like:

`const actual = buildImageURL("ABC123", "IIIF_SQUARE");` ?

assets/js/services/helpers.test.js
```diff
@@ -0,0 +1,75 @@
...
+  it("should return is-success & is-light for VALID", () => {
+    const expected = "is-success is-light";
+    const actual = getClassFromIngestSheetStatus("VALID");
+    expect(expected).toEqual(actual);
+  });
+  it("should return empty", () => {
+    const actual = getClassFromIngestSheetStatus();
+    expect(actual).toBe("");
+  });
+});
```
adamjarling: Looks great!

adamjarling:
@katdivyareddy10 I just merged in a branch, so you'll need to rebase this branch.  You might get merge conflicts.  If you need any help, or want to walk through it ping me and we can screen share.

assets/js/components/Work/ListItem.jsx
katdivyareddy10: If fileset didn't exist, this would fail like "Cannot read property length of undef". I thought about the length and remembered that error. Let me know if you still want me to change this.

assets/js/services/helpers.js
katdivyareddy10: Sounds good! I'll do that!

assets/js/services/helpers.test.js
katdivyareddy10: Cool, I'll make this change!

assets/js/services/helpers.test.js
katdivyareddy10: Ok, I'll change it to reflect close to real value.

assets/js/services/helpers.test.js
katdivyareddy10: Thank you!

katdivyareddy10:
> @katdivyareddy10 I just merged in a branch, so you'll need to rebase this branch. You might get merge conflicts. If you need any help, or want to walk through it ping me and we can screen share.

Let me try the rebase once and see if I can do it myself and if I run into issues, I'll ping you.

assets/js/components/Work/ListItem.jsx
katdivyareddy10:
``` 
var work = {};
undefined
work.filesets.length
VM245:1 Uncaught TypeError: Cannot read property 'length' of undefined
    at <anonymous>:1:15
```
Something like this.

adamjarling: Ah, good point.  Make sure you're testing `fileSets` and not `filesets`.   If that's fine, then yea, include the check for `work.fileSets` but I still like the length check better than checking for a `[0]` element.

adamjarling:
@katdivyareddy10 I'm guessing the conflict in `global-vars.js` is because we both added to the file.  If that's the case, you'll want to resolve the conflict by accepting both changes.... yours and mine.

PR #547, opened 2020-02-26
Add react-hook-form package and general patterns for how to handle Gr…

adamjarling:
…aphQL mutations from Work tabs

This PR uses the first Work tab "About" as an example of how to set up forms.  It pulls in the package `react-hook-form` which gives us:
- Validation
- Cleaner implementation of form element markup (less code)
- Dirty checking

This PR gives an example of wiring up GraphQL `updateWork` mutation for Work tabs.

Finally, it provides an example pattern of testing each Work tab.

image

Lastly, lastly, in preparation for the demo I put forward a less striking way of displaying items in need of API support.   This is an attempt to merry existing API support with wireframe fields.

image


assets/js/components/Work/Tabs/About.jsx
```diff
@@ -1,134 +1,211 @@
...
-                    type="date"
-                    value={insertedAt}
-                    onChange={e => setInsertedAt(e.target.value)}
-                  />
-                ) : (
-                  <p>{formatDate(work.insertedAt)}</p>
-                )}
+              <div className="field">
+                <label className="label">
+                  Date Created{" "}
```
kdid: Date Created is not the same thing as `inserted_at`
`inserted_at` is an autogenerated database field (generated when the work is created) and is not editable. (There is an `updated_at` field that is auto generated when the record is updated)

Date Created is a field of descriptive metadata that is not yet exposed in the API. It is an EDTF date, so a regular calendar would not be appropriate as a chooser. (It will also eventually be a multi-valued field)

kdid:
Overall looks like a good pattern - left a comment about the date.

assets/js/components/Work/Tabs/About.jsx
adamjarling: Ah, good to know.  Thanks for the info @kdid

adamjarling:
Just added a patch to this commit which allows "Edit" state to persist across tab navigation.  Before it was resetting back to a "View" state when navigating between Work tabs, which could be annoying if a user had edits in progress (which were getting wiped out).

PR #553, opened 2020-02-27
Fix the hiccup of updating UI post work creation, plus UI cleanup to …

adamjarling:
…IngestSheet component tree

Thanks @kdid for pointing me in the right direction for original intent of this ticket.  I took that as a jumping-off point to do some other UI cleanup work which had been bugging me for a while.  This PR brings up the primary action buttons for Ingest Sheet screen to the "hero section", to keep consistent with primary action button placement in rest of the application.

Also update the breadcrumbs to make the visual a little more subtle.  Maybe?  

image


adamjarling:
Good catch @katdivyareddy10 Yea, that alert needs to be wrapped by a `container` class.

adamjarling:
@katdivyareddy10 Since I did an update to the original commit and force pushed, I can't seem to find how to resolve the request for changes.  Mind having another look at this, and if you approve mark as Approved?

katdivyareddy10:
Just 1 minor fix, I believe the approve ingest sheet notification needs to be wrapped too. Screenshot for reference. 
Screen Shot 2020-03-02 at 11 55 46 AM

PR #564, opened 2020-03-03
Added edit form for structure tab

katdivyareddy10:
Form to edit structure tab accession number & description fields is setup.

Screen Shot 2020-03-03 at 12 34 53 PM

Added new changes as per PR suggestions and made the accession number to readonly and added label as edit input.

Screen Shot 2020-03-04 at 11 08 17 AM

### DIFF
--- a/assets/js/components/Work/Tabs/Structure.jsx
+++ b/assets/js/components/Work/Tabs/Structure.jsx
@@ -54,7 +54,7 @@ const WorkTabsStructure = ({ work }) => {
                                 <input
                                   ref={register({ required: true })}
                                   name={`label-${id}`}
-                                  data-testid="label"
+                                  data-testid="input-label"
                                   className={`input ${
                                     errors[`label-${id}`] ? "is-danger" : ""
                                   }`}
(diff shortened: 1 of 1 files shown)


kdid:
We don't allow accession edits. (We would eventually prevent this on the backend.)

assets/js/components/Work/Tabs/Structure.jsx
```diff
@@ -2,74 +2,190 @@ import React, { useContext } from "react";
...
+                          <div class="field-label is-normal">
+                            <label class="label">Description</label>
+                          </div>
+                          <div class="field-body">
+                            <div class="field">
+                              <p class="control">
+                                <textarea
+                                  ref={register({ required: true })}
+                                  name={`metadataDescription-${id}`}
+                                  data-testid="metadataDescription"
```
adamjarling: I made a comment in another PR, and what I'd like to do for consistency in our code is handle `data-testid`s as follows:

`data-testid="[element type]-[name of thing separated by dashes]`

So, this test id would change to:
`data-testid="metadataDescription"` becomes `data-testid="textarea-metadata-description"`

And apply this to all `data-testid`s in this form please.

PR #565, opened 2020-03-03
Index FileSets on Work document in Elasticsearch

bmquinn:
FileSets get indexed into works like this: 
```elixir
file_sets: [
    %{
      accession_number: "PwyKWICT",
      id: "b7917457-6904-412f-8271-a3a68e5b1de1",
      label: "test"
    }
]
```

### DIFF
--- a/lib/meadow/data/schemas/work.ex
+++ b/lib/meadow/data/schemas/work.ex
@@ -78,6 +78,15 @@ defmodule Meadow.Data.Schemas.Work do
             nil -> %{}
             collection -> %{id: collection.id, title: collection.name}
           end,
+        file_sets:
+          work.file_sets
+          |> Enum.map(fn file_set ->
+            %{
+              id: file_set.id,
+              accession_number: file_set.accession_number,
+              label: file_set.metadata.label
+            }
+          end),
         create_date: work.inserted_at,
         iiif_manifest: IIIF.manifest_id(work.id),
         modified_date: work.updated_at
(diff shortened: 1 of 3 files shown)


lib/meadow/data/schemas/work.ex
```diff
@@ -77,6 +77,15 @@ defmodule Meadow.Data.Schemas.Work do
             nil -> %{}
             collection -> %{id: collection.id, title: collection.name}
           end,
+        file_sets:
+          work.file_sets
+          |> Enum.map(fn file_set ->
+            %{
+              id: file_set.id,
+              accession_number: file_set.accession_number,
+              label: file_set.metadata.description
```
kdid: The label is a actually the label. But I think my Manifest PR adds it in....
The label is populated with original_filename on ingest and then is editable.

Should we just wait to merge mine in and then you can rebase?

lib/meadow/elasticsearch_diff_store.ex
```diff
@@ -30,7 +30,7 @@ defmodule Meadow.ElasticsearchDiffStore do
     |> Repo.stream()
     |> Stream.chunk_every(@chunk_size)
     |> Stream.flat_map(fn chunk ->
-      Repo.preload(chunk, :collection)
+      Repo.preload(chunk, [:collection, :file_sets])
```
kdid: Is this efficient? I'm only asking because I started to do this yesterday and then started googling and it seemed like it was suggested to manually join over doing this. But I trust you if you've looked into it.

bmquinn: When I was reading up on this I saw a Twitter thread by Michał Muskała that hopefully clarifies it a bit: https://twitter.com/michalmuskala/status/1118119934758928385.

PR #567, opened 2020-03-04
Modified collectionForm to use React-hook-form

katdivyareddy10:
Changed collectionForm component to use react-hook-form.
Edit/Add new collection works as expected.
Screen Shot 2020-03-04 at 8 43 58 AM

### DIFF
--- a/assets/js/components/Collection/Form.jsx
+++ b/assets/js/components/Collection/Form.jsx
@@ -75,7 +75,7 @@ const CollectionForm = ({ collection }) => {
         <div className="columns is-centered">
           <div className="column is-half">
             <div className="field">
-              <label htmlFor="collectionName" className="label">
+              <label htmlFor="collection-name" className="label">
                 Collection Name
               </label>
               <div className="control">
@@ -85,9 +85,9 @@ const CollectionForm = ({ collection }) => {
                   type="text"
                   ref={register({ required: true })}
                   name="collectionName"
-                  id="colletion-name"
+                  id="collection-name"
                   defaultValue={collection ? collection.name : ""}
-                  data-testid="collection-name"
+                  data-testid="input-collection-name"
                 />
               </div>
             </div>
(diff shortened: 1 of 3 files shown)


assets/js/components/Collection/Form.jsx
```diff
@@ -83,64 +57,53 @@ const CollectionForm = ({ collection }) => {
...
-              type="text"
-              name="collectionName"
-              id="collectionName"
-              label="Collection Name"
-              value={formValues.collectionName || ""}
-              onChange={handleInputChange}
-              data-testid="collection-name"
-            />
+            <div className="field">
+              <label htmlFor="collectionName" className="label">
```
adamjarling: This might have been my bad initially setting up, but the `aria` `htmlFor`attribute targets an `id`, not a `name` attribute.  Can you update the `<label>` elements everywhere in this form to look like:

```
<label htmlFor="[id of element]"...
```

assets/js/components/Collection/Form.jsx
```diff
@@ -83,64 +57,53 @@ const CollectionForm = ({ collection }) => {
...
+                Collection Name
+              </label>
+              <div className="control">
+                <input
+                  placeholder="Add collection Name"
+                  className={`input ${errors.collectioName ? "is-danger" : ""}`}
+                  type="text"
+                  ref={register({ required: true })}
+                  name="collectionName"
+                  id="colletion-name"
```
adamjarling: Minor typo. `colletion-name` should be `collection-name`

assets/js/components/Collection/Form.jsx
```diff
@@ -83,64 +57,53 @@ const CollectionForm = ({ collection }) => {
...
+                  defaultValue={collection ? collection.name : ""}
+                  data-testid="collection-name"
```
adamjarling: How about we establish a pattern to help us in advance as we start increasing test coverage.   For naming `data-testid`s, if it's a form element, let's use this convention as a rule of thumb:

`data-testid="[type of form element - name of form element`

So `data-testid="collection-name"` becomes `data-testid="input-collection-name"`

... and so forth in all the form elements which have test ids.

Thoughts?

PR #573, opened 2020-03-04
Add addWorksToCollection and removeWorksFromCollection mutations

mbklein:
- Add `add_works/2` and `remove_works/2` to `Meadow.Data.Collections`

We should consider removing `addWorkToCollection` during the bi-fortnight since `addWorksToCollection` is more flexible.

### DIFF
--- a/lib/meadow/data/collections.ex
+++ b/lib/meadow/data/collections.ex
@@ -5,7 +5,7 @@ defmodule Meadow.Data.Collections do
 
   import Ecto.Changeset
   import Ecto.Query, warn: false
-  alias Meadow.Data.Schemas.Collection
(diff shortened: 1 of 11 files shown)


kdid:
@mbklein - this looks good. Did you decide not to do the collection add/change functionality in the  `updateWork` mutation?

bmquinn:
Looks good, but I have a tiny question: since `update_all/3` doesn't update autogenerated columns like `updated_at` (see https://hexdocs.pm/ecto/Ecto.Repo.html#c:update_all/3). Assuming the `notify_work_changes()` trigger still works for Elasticsearch indexing (haven't checked locally yet), do we care about not updating autogenerated fields?

PR #574, opened 2020-03-04
Add project and sheet to works in GraphQL

kdid:
* Exposed project and ingest sheet on the work object in GraphQL
* Added `Sheets.works_with_sheets/1` to works queries (except sheet-works query - so requesting project/sheet info in `ingestSheetWorks` query will cause error. Out of scope for this PR. We'll open issue... We may not need to do this if we are all aware...)
* Displayed on  the Administrative tab 

Screen Shot 2020-03-04 at 3 41 31 PM

### DIFF
--- a/assets/js/components/Work/work.query.js
+++ b/assets/js/components/Work/work.query.js
@@ -23,6 +23,14 @@ export const GET_WORK = gql`
       }
       insertedAt
       manifestUrl
+      project {
+        id
+        name
+      }
+      sheet {
+        id
+        name
+      }
       published
       representativeImage
       updatedAt
@@ -63,6 +71,14 @@ export const GET_WORKS = gql`
       }
       insertedAt
       manifestUrl
+      project {
+        id
+        name
+      }
+      sheet {
+        id
+        name
+      }
       published
       representativeImage
       updatedAt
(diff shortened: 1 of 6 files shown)


mbklein:
So the upshot is that all `Work`s loaded from within the `Works` context will _always_ have their project/sheet info loaded with them now?

kdid:
I mean our thought was basically, yes? Do you think that's an issue?

bmquinn:
It was either that or make new `Works` queries and tests and update all the `Resolver` functions to use them. We talked about the queries being essentially there to power the React app for the foreseeable future, so might as well just load up the works with their ingest sheets/projects all the time until we need works without that "extra" information. Happy to be persuaded otherwise, though 😄

kdid:
To add to what Brendan said - all the graphQL queries return the same type of object so they should support the same fields.. . right? (That's why the ingest_sheet_works query doesn't work.)

assets/js/components/Work/work.query.js
```diff
@@ -23,6 +23,14 @@ export const GET_WORK = gql`
...
       manifestUrl
+      project {
+        id
+        name
+      }
+      sheet {
+        id
+        name
+      }
       published
```
adamjarling: Yea, this is perfect and all it needs.

PR #575, opened 2020-03-04
Fix bug in publish flag

kdid:
Publish flag was not being `cast` in the changeset and so updating it through the `updateWork` mutation wasn't working. 

I created a separate `update_changeset` since we don't have the same requirements for updating fields as creating fields. (can't edit accession_number, required fields not the same, etc...)


adamjarling:
I'm going to add a commit to this which updates the front-end...

PR #581, opened 2020-03-05
Wire up setWorkImage UI in Work Structure tab

adamjarling:
Partial PR for https://github.com/nulib/repodev_planning_and_docs/issues/556

This feature happens asynchronously, outside of "Edit" mode.   It disallows the ability to have no representative image after one has been used also... assuming we want that @davidschober ?  

image


davidschober:
@adamjarling Yes, I think that makes sense. A work always has to have a representative image. On ingest we should set the default to the first in the set of images. But a user should never be able to set a work to have no representative image.

PR #582, opened 2020-03-05
Work delete button is wired up and confirmation is toasted

katdivyareddy10:
Work delete button all wired up
Added fixes to the breadcrumb as mentioned in issue 570

Screen Shot 2020-03-05 at 11 19 37 AM

Changes as per comments to show accession number if title is empty

Screen Shot 2020-03-05 at 1 08 15 PM

### DIFF
--- a/assets/js/components/Work/work.query.js
+++ b/assets/js/components/Work/work.query.js
@@ -118,3 +118,22 @@ export const ADD_WORK_TO_COLLECTION = gql`
     }
   }
 `;
+
+export const DELETE_WORK = gql`
+  mutation deleteWork($workId: ID!) {
+    deleteWork(workId: $workId) {
+      id
+      descriptiveMetadata {
+        title
+      }
+      sheet {
+        id
+        name
+      }
+      project {
+        id
+        name
+      }
+    }
+  }
+`;
--- a/assets/js/screens/Work/Work.jsx
+++ b/assets/js/screens/Work/Work.jsx
@@ -1,6 +1,12 @@
-import React from "react";
+import React, { useState } from "react";
 import { useMutation, useQuery } from "@apollo/react-hooks";
-import { GET_WORK, UPDATE_WORK } from "../../components/Work/work.query";
+import {
+  GET_WORK,
+  UPDATE_WORK,
+  DELETE_WORK
+} from "../../components/Work/work.query";
+import UIModalDelete from "../../components/UI/Modal/Delete";
+import { useHistory } from "react-router-dom";
 import { useParams } from "react-router-dom";
 import Layout from "../Layout";
 import Error from "../../components/UI/Error";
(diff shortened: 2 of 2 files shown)


katdivyareddy10:
This PR also contains code for issue 570

assets/js/screens/Work/Work.jsx
```diff
@@ -85,6 +120,12 @@ const ScreensWork = () => {
       </section>
       <UIBreadcrumbs items={breadCrumbs} data-testid="work-breadcrumbs" />
       <Work work={data.work} />
+      <UIModalDelete
+        isOpen={deleteModalOpen}
+        handleClose={onCloseModal}
+        handleConfirm={handleDeleteClick}
+        thingToDeleteLabel={`Work ${descriptiveMetadata.title || ""}`}
```
adamjarling: Can you add some type of check here in case of `descriptiveMetadata.title` doesn't exist?   Currently it says "Work null successfully deleted".  How about we not include "null" in the message?   Just "Work deleted" or something....

assets/js/components/Work/work.query.js
```diff
@@ -118,3 +118,22 @@ export const ADD_WORK_TO_COLLECTION = gql`
     }
   }
 `;
+
+export const DELETE_WORK = gql`
+  mutation deleteWork($workId: ID!) {
+    deleteWork(workId: $workId) {
+      id
```
adamjarling: Just checking if we really need to return all this data beyond `id`?   It might make for more to maintain... Unless Apollo cache needs it to make it's own update, but I don't think it does....   Can you experiment removing the extra data and see what happens?

katdivyareddy10: I started with ID initially and realized I need to return more data for the navigation to get back to the ingest sheet.

assets/js/screens/Work/Work.jsx
katdivyareddy10: I added "|| " to fix that null. Are you still getting it when deleting a work with no title?

kdid: Could we maybe do something like show the accession number instead if there is no title? No works have titles on ingest. Works always have accession numbers...

adamjarling: Yea, that might be the way to go.

katdivyareddy10: I thought about it but i expected if there's a field for title, they must be using it. I need to stop assuming :D , I'll add this and adam's suggestion as well.

assets/js/screens/Work/Work.jsx
```diff
@@ -85,6 +120,16 @@ const ScreensWork = () => {
...
       <Work work={data.work} />
+      <UIModalDelete
+        isOpen={deleteModalOpen}
+        handleClose={onCloseModal}
+        handleConfirm={handleDeleteClick}
+        thingToDeleteLabel={`Work ${
+          descriptiveMetadata
+            ? descriptiveMetadata.title || accessionNumber
+            : accessionNumber
+        }`}
```
adamjarling: Nice!   Getting fancy on the logic.

PR #585, opened 2020-03-05
Add ADR and mix task to help with semantic versioning and tagging

mbklein:
In light of how far Meadow has come while still retaining the same hard-coded version number, @bmquinn and I were discussing when and how to update it. This draft PR includes a proposed ADR and mix task to deal with it in a relatively simple fashion, which will also help when it's time to implement https://github.com/nulib/repodev_planning_and_docs/issues/575

### DIFF
--- a/doc/architecture/decisions/0024-version-management.md
+++ b/doc/architecture/decisions/0024-version-management.md
@@ -1,0 +1,28 @@
+# 24. Version Management
+
+Date: 2020-03-05
+
+## Status
+
+Accepted
+
+## Context
+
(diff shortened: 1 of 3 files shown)


kdid:
I'm not clear on why we would version Meadow? It isn't used by anyone else and doesn't have a public API. Even if we considered the GraphQL API public, I don't think you version GraphQL API's in the same way -  I think you just deprecate and introduce fields in a fluid way.

bmquinn:
I think versioning makes sense even for internal communication purposes. Like "v1.0" means we've decided on a team that we've hit the stability and feature set we need for going to production and not making breaking changes without discussion.

PR #663, opened 2020-04-07
Added mobile navigation links and modified code to reuse existing links

katdivyareddy10:
Mobile links are wired up and navbar links are cleaned up 

Screen Shot 2020-04-07 at 5 34 49 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ See your local "s3" buckets.
 
 ### GraphQL API
 
-You can visit the GraphiQL interface at: [`http://devbox.library.northwestern.edu/api/graphiql`](http://devbox.library.northwestern.edu/api/graphiql)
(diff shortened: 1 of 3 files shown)


katdivyareddy10:
The links are open by default in Bulma and there's no class to close/open them. I can add a custom function to determine the display hidden/show. Let me know what you think.

adamjarling:
Ah, gotcha.  Let's see if we can implement a React solution for this?   Check out this Codepen for some possible ideas?    https://codepen.io/Nikitoss334/pen/VOEdVY

katdivyareddy10:
I used the checkbox idea from the CSS only Codepen @adamjarling  provided for reference. 
Screen Shot 2020-04-08 at 2 12 45 PM

adamjarling:
@katdivyareddy10 This looks fantastic.  Thanks for the update!  One last change (I promise then I think we're good here).

Can you take what you learned from `aria` accessibility, and make the dropdown elements are best-practice, aria labeling compliant?

katdivyareddy10:
I added aria labels, roles and aria-expanded items to the navbar.

PR #687, opened 2020-04-14
Add tests for collections

katdivyareddy10:
Jest tests added for collection screen, collection component and list item.

Screen Shot 2020-04-14 at 1 40 40 PM

### DIFF
--- a/assets/js/components/Collection/Collection.test.jsx
+++ b/assets/js/components/Collection/Collection.test.jsx
@@ -1,49 +1,33 @@
 import React from "react";
+import { render } from "@testing-library/react";
+import { wait, waitForElement } from "@testing-library/react";
 import Collection from "./Collection";
-import { GET_COLLECTION } from "../../components/Collection/collection.query";
-import { renderWithRouterApollo } from "../../services/testing-helpers";
-import { Route } from "react-router-dom";
-import { wait } from "@testing-library/react";
 
-const mocks = [
-  {
-    request: {
-      query: GET_COLLECTION,
-      variables: {
-        id: "7a6c7b35-41a6-465a-9be2-0587c6b39ae0"
-      }
-    },
-    result: {
-      data: {
-        collection: {
-          adminEmail: "test@test.com",
-          description: "Test arrays keyword arrays arrays arrays arrays",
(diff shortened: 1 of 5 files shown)


assets/js/components/Collection/Collection.test.jsx
```diff
@@ -1,10 +1,49 @@
 import React from "react";
```
adamjarling, line 1: @katdivyareddy10 Can we add some more tests here which validate that props are passing through and displaying (ie. Description, Admin Email) in the component?

And also maybe test that the Collection image (or placeholder at least) is present.   And that the component displays the `CollectionSearch` component?

assets/js/screens/Collection/List.test.js
```diff
@@ -0,0 +1,47 @@
+import React from "react";
```
adamjarling, line 1: For this component test, seems like the component itself does the following:

- Renders out "hero" (this will change to "header" or something similar with the new designs, but easy enough to update later) content.
- Renders out a "Create" button.
- Renders out a collection list section
- Renders out individual list items (`CollectionListRow`)

Can you add some extra tests to cover the general functionality above, if any are missing?

assets/js/screens/Collection/Collection.test.js
```diff
@@ -0,0 +1,67 @@
+import React from "react";
```
adamjarling, line 1: Same here, can you add some extra tests which cover:

- Rendering of Edit / Delete buttons.
- Maybe even a test for clicking on each button, what should happen?   Is it possible to test?

assets/js/screens/Collection/List.test.js
katdivyareddy10: I already renders collection list and hero sections, I added create button and list item tests.

assets/js/screens/Collection/Collection.test.js
katdivyareddy10: I was able to test delete button click fully. For edit button, the click event results in firing nested un-existing mock events due to the way it is rendered with RouterApollo. I was able to test that edit button exists on screen. The latter part will be covered in integration testing.

PR #720, opened 2020-04-24
UI sweep through cleanup

adamjarling:
This PR is another iterative design revision touching a bunch of stuff.  Increases consistency, and some minor improvements.

### DIFF
--- a/assets/js/screens/IngestSheet/IngestSheet.jsx
+++ b/assets/js/screens/IngestSheet/IngestSheet.jsx
@@ -1,18 +1,18 @@
 import React from "react";
-import { withRouter } from "react-router-dom";
 import Error from "../../components/UI/Error";
 import UILoadingPage from "../../components/UI/LoadingPage";
+import UILoading from "../../components/UI/Loading";
 import IngestSheet from "../../components/IngestSheet/IngestSheet";
 import gql from "graphql-tag";
 import { useQuery } from "@apollo/react-hooks";
 import {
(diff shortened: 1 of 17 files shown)


assets/js/screens/IngestSheet/IngestSheet.jsx
```diff
@@ -1,18 +1,18 @@
 import React from "react";
-import { withRouter } from "react-router-dom";
 import Error from "../../components/UI/Error";
 import UILoadingPage from "../../components/UI/LoadingPage";
```
katdivyareddy10, line 3: This is not being used in this component, can be removed.

assets/styles/scss/_variables.scss
```diff
@@ -52,7 +52,7 @@ $CamptonExtraLight: "Campton Extra Light", "Courier New", sans-serif;
 // Colors
 $primary: $nu-purple;
 $link: $nu-purple;
-$info: $blue;
+$info: $rich-black-80;
```
katdivyareddy10, line 55: Is there a reason this was changed to black ? I tried to ingest a sheet, the colors don't seem to be in sync on the page.
Screen Shot 2020-04-24 at 4 01 49 PM

adamjarling: Yea, I think the blue just doesn't make sense maybe?  I've noticed in some other Northwestern emails, that they use the NU "rich black" color to highlight information, so was playing around with that idea I suppose.  This will probably change about 5 more times, just experimenting, not a final decision.

PR #732, opened 2020-04-29
Mock response object for simple controlled vocabularies in work related queries

kdid:
- Partial PR for https://github.com/nulib/repodev_planning_and_docs/issues/691
- Mocks responses for simple controlled vocabulary fields in works queries using client only fields in the Apollo client as described here. (Mock returns an array of 0-4 id/label pairs) https://www.apollographql.com/docs/react/development-testing/client-schema-mocking/
- You can query for them in the Apollo client browser extension's GraphiQL interface or use them in the app, but won't see responses in the real GraphQLPlayground.
- added `@desc NOT YET IMPLEMENTED` comments to the properties in the Absinthe schema - which will be viewable in GraphQL Playground.
- It looks like you need to include the __typename within that Apollo extension's graphiQL for the queries to work. ex:
genre {
  id
  label
  __typename
}
```

Simple fields are listed here: http://docs.rdc.library.northwestern.edu/2._Developer_Guides/Donut/Controlled-Vocabularies/

### CODE REVIEW

File: assets/js/services/testing-helpers.js [hunk 1]
```diff
@@ -50,7 +51,7 @@ export function renderWithRouterApollo(
   } = {}
 ) {
   const Wrapper = ({ children }) => (
-    <MockedProvider mocks={mocks} addTypename={false}>
+    <MockedProvider mocks={mocks} addTypename={false} resolvers={resolvers}>
```
adamjarling, line 54: @kdid So this makes it work with tests eh?   Nice.

kdid: It was throwing an error until I googled and added that. Not sure i totally understand how that's all working.

PR #734, opened 2020-04-29
Structure out metadata editing forms to hopefully scale better

adamjarling:
Partial PR for cleaning up Metadata form tabs.

image

### DIFF
--- a/assets/js/components/Work/Tabs/Administrative.jsx
+++ b/assets/js/components/Work/Tabs/Administrative.jsx
@@ -6,7 +6,7 @@ import { toastWrapper } from "../../../services/helpers";
 import {
   VISIBILITY_OPTIONS,
   RIGHTS_STATEMENTS,
-  PRESERVATION_LEVELS
+  PRESERVATION_LEVELS,
 } from "../../../services/global-vars";
 import { GET_COLLECTIONS } from "../../Collection/collection.query";
 import { UPDATE_WORK, ADD_WORK_TO_COLLECTION, GET_WORK } from "../work.query";
@@ -14,15 +14,18 @@ import { useForm } from "react-hook-form";
 import UITagNotYetSupported from "../../UI/TagNotYetSupported";
 import { Link } from "react-router-dom";
 import { setVisibilityClass } from "../../../services/helpers";
+import UIFormSelect from "../../UI/Form/Select";
+import UIFormField from "../../UI/Form/Field";
+import WorkTabsHeader from "./Header";
 
 const WorkTabsAdministrative = ({ work }) => {
   const { id, administrativeMetadata, collection, project, sheet } = work;
   const [isEditing, setIsEditing] = useIsEditing();
-  const { register, handleSubmit } = useForm();
+  const { register, handleSubmit, errors } = useForm();
   const [updateWork] = useMutation(UPDATE_WORK, {
     onCompleted({ updateWork }) {
       setIsEditing(false);
-    }
+    },
   });
 
   // TODO: Add Work to collection is disrupting changes made using updateWork,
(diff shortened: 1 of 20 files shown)


assets/js/components/Work/Tabs/Structure.jsx
```diff
@@ -25,222 +30,179 @@ const WorkTabsStructure = ({ work }) => {
...
+          </>
+        )}
+      </WorkTabsHeader>
+
+      <div style={{ marginTop: "1rem" }}>
+        {work.fileSets.map(({ id, accessionNumber, metadata }) => (
+          <article key={id} className="box">
+            <div className="columns">
+              <div className="column is-2">
+                <figure className="image">
```
katdivyareddy10, line 89: Can you add a class to this image so when the screen becomes smaller, the image size is also reduced ? It is looking rather big compared to the content that is being edited. 
Screen Shot 2020-04-29 at 2 47 45 PM

assets/js/screens/Work/Work.jsx
```diff
@@ -59,31 +59,39 @@ const ScreensWork = () => {
...
+      route: `/project/list`,
+    },
     {
       label: `${project.name}`,
-      route: `/project/${project.id}`
+      route: `/project/${project.id}`,
+    },
+    {
+      label: "Ingest Sheets",
+      route: `/project/${id}`,
```
katdivyareddy10, line 76: This is passing work ID and not project ID as expected.

Screen Shot 2020-04-29 at 3 20 59 PM

assets/js/components/Work/Tabs/Administrative.jsx
```diff
@@ -34,190 +37,158 @@ const WorkTabsAdministrative = ({ work }) => {
...
+        rightsStatement: data.rightsStatement,
       },
       visibility: data.visibility,
 
-      published: true
+      published: true,
     };
     updateWork({
-      variables: { id, work: workUpdateInput }
+      variables: { id, work: workUpdateInput },
```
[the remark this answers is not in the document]
adamjarling: @katdivyareddy10 Are they updating on `deploy/staging`?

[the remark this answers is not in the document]
adamjarling: @kdid @bmquinn  Mind sanity checking this for us?   Seems like we're getting the right response from the API mutation in the app and in GraphQL Playground, but the data doesn't seem to persist.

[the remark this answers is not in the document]
adamjarling: Ok, I think the cache is not updating.  It wasn't working, but now my data magically appeared.  Looking into it.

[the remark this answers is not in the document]
adamjarling: @katdivyareddy10 This is in `deploy/staging` as well, and is happening (I'm pretty sure) because we're calling 2 update `mutation`s in the `onSubmit()` function.  This is messing with the cache.  I'll create a new issue for it.   Not related to this PR.

assets/js/components/Work/Tabs/Structure.jsx
adamjarling: For this, mobile views are fairly minimal and how about we wait until someone says something negative about it?  It's kinda tricky to block everything out fullwidth on mobile and have the image look decent without adding extra code bloat (for now) to make it smaller, centered, and such.   Or I'm not sure of a good solution at the current moment.  I agree not ideal, but maybe best case of not that great a selection of cases?

assets/js/screens/Work/Work.jsx
adamjarling: So, the logic is, though it might not be good logic, is that "Ingest Sheets" refers to a list of Ingest Sheets, which technically live on an individual Project screen.   Make sense?

Agreed we need a good/better breadcrumb strategy for nested content to make it obvious to users where they're at in the app, and ideally reinforce the association between Projects, Ingest Sheets, Works, and Filesets.

I'll kick off a ticket and get @davidschober 's thoughts.

katdivyareddy10: Ok, I think we can remove it from this coz it is throwing info not found error on the page. If you go into a work and click on "Ingest sheets", it should reproduce that error.

PR #744, opened 2020-04-30
Add remaining uncontrolled fields to WorkDescriptiveMetadata

mbklein:
Updates `WorkDescriptiveMetadata` and the associated GraphQL types to match the field names / repeatability in the [metadata model](https://docs.google.com/spreadsheets/d/1PBvp9FJAvG6JIP6RZwmiZSDhgYLQ6VRSknUnkodEH8Q/edit#gid=396400352)

### DIFF
--- a/lib/meadow/data/schemas/work.ex
+++ b/lib/meadow/data/schemas/work.ex
@@ -80,9 +80,7 @@ defmodule Meadow.Data.Schemas.Work do
     def encode(work) do
       %{
         model: %{application: "Meadow", name: String.capitalize(work.work_type)},
-        title: work.descriptive_metadata.title,
         accession_number: work.accession_number,
-        description: work.descriptive_metadata.description,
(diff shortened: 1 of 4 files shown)


mbklein:
I really wanted to make [this list](https://github.com/nulib/meadow/pull/744/files#diff-9745bd4a4a98647a0c94e4a81cd30e07R121) dynamic the same way I did [this one](https://github.com/nulib/meadow/pull/744/files#diff-904bd6c22a509329772f71435ead3fa5R43), but something about Absinthe's metaprogramming/macro structure wouldn't let me.

kdid:
I'm thinking we should remove the `ark` from the chunk of uncontrolled fields and instead adding it manually to the `WorkDescriptiveMetadata` since it shouldn't be in the `WorkDescriptiveMetadataInput ` since we will eventually be generating arks.

PR #775, opened 2020-05-06
Changes to search page to use work components toggler

katdivyareddy10:
Search page uses toggle view with `workcard` and `worklist` components.
Tests have been modified after rebasing with deploy/staging due to changes in search results and work component structure.

Screen Shot 2020-05-06 at 3 46 31 PM
Screen Shot 2020-05-06 at 3 49 59 PM
Screen Shot 2020-05-06 at 4 48 48 PM

### DIFF
--- a/assets/js/components/Work/UIWorkCardItem.jsx
+++ b/assets/js/components/Work/UIWorkCardItem.jsx
@@ -44,7 +44,7 @@ const WorkCardItem = ({
 
         <div className="content">
           <p>
-            <span className="tag">{workType.label}</span>
+            <span className="tag">{workType.label.toUpperCase()}</span>
             {visibility && (
               <span
                 data-testid="tag-visibility"
--- a/assets/js/components/Work/UIWorkCardItem.jsx
+++ b/assets/js/components/Work/UIWorkCardItem.jsx
@@ -1,45 +1,59 @@
-import React from "react";
+import React, { useContext } from "react";
 import PropTypes from "prop-types";
 import { Link } from "react-router-dom";
 import { setVisibilityClass, formatDate } from "../../services/helpers";
+import { IIIFProvider, IIIFContext } from "../../components/IIIF/IIIFProvider";
 
-const WorkCardItem = ({ work }) => {
+const WorkCardItem = ({
+  id,
+  representativeImage,
+  title,
+  descriptiveMetadata,
+  workType,
+  visibility,
+  published,
+  accessionNumber,
+  fileSets,
+  manifestUrl,
+  updatedAt,
+}) => {
+  const iiifServerUrl = useContext(IIIFContext);
   return (
     <div className="card " data-testid="ui-workcard">
       <div className="card-image">
         <figure className="image is-4by3">
-          <Link to={`/work/${work.id}`}>
+          <Link to={`/work/${id}`}>
             <img
               src={`${
-                work.representativeImage
-                  ? work.representativeImage + "/full/1280,960/0/default.jpg"
-                  : "/images/1280x960.png"
+                representativeImage.id
+                  ? iiifServerUrl +
+                    representativeImage.id +
+                    "/square/500,500/0/default.jpg"
+                  : representativeImage + "/full/1280,960/0/default.jpg"
               }`}
               data-testid="image-work"
-              alt={work.title}
+              alt={title}
             />
           </Link>
         </figure>
       </div>
       <div className="card-content">
         <h3 className="title is-size-4">
-          {work.descriptiveMetadata.title
-            ? work.descriptiveMetadata.title
-            : "Untitled"}
+          {descriptiveMetadata.title ? descriptiveMetadata.title : "Untitled"}
         </h3>
 
         <div className="content">
           <p>
-            <span className="tag">{work.workType && work.workType.label}</span>
-            {work.visibility && (
+            <span className="tag">{workType.label}</span>
+            {visibility && (
               <span
                 data-testid="tag-visibility"
-                className={`tag ${setVisibilityClass(work.visibility.id)}`}
+                className={`tag ${setVisibilityClass(visibility.id)}`}
               >
-                {work.visibility.label.toUpperCase()}
+                {visibility.label.toUpperCase()}
               </span>
             )}
-            {work.published && (
+            {published && (
               <span data-testid="dd-published" className="tag is-success">
                 PUBLISHED
               </span>
@@ -47,18 +61,18 @@ const WorkCardItem = ({ work }) => {
           </p>
           <dl>
             <dt>Accession Number:</dt>
-            <dd data-testid="dd-accession-number">{work.accessionNumber}</dd>
+            <dd data-testid="dd-accession-number">{accessionNumber}</dd>
             <dt>Filesets:</dt>
             <dd>
               <span className="tag is-light" data-testid="dd-filesets-length">
-                {work.fileSets.length}
+                {fileSets.length}
               </span>
             </dd>
             <dt>Last Updated: </dt>
-            <dd data-testid="dd-updated-date">{formatDate(work.updatedAt)}</dd>
+            <dd data-testid="dd-updated-date">{formatDate(updatedAt)}</dd>
             <dt>IIIF Manifest:</dt>
             <dd>
-              <a href={work.manifestUrl} target="_blank">
+              <a href={manifestUrl} target="_blank">
                 <u>JSON File</u>
               </a>
             </dd>
--- a/assets/js/components/Work/UIWorkCardItem.jsx
+++ b/assets/js/components/Work/UIWorkCardItem.jsx
@@ -2,13 +2,11 @@ import React, { useContext } from "react";
 import PropTypes from "prop-types";
 import { Link } from "react-router-dom";
 import { setVisibilityClass, formatDate } from "../../services/helpers";
-import { IIIFProvider, IIIFContext } from "../../components/IIIF/IIIFProvider";
 
 const WorkCardItem = ({
   id,
   representativeImage,
   title,
-  descriptiveMetadata,
   workType,
   visibility,
   published,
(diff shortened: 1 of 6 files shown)


adamjarling:
@katdivyareddy10 I think you'll need to rebase this on `deploy/staging`...before re-triggering a Circle CI build.  I snuck in a merge last night before it, and Karen might have a few in as well.

assets/js/screens/Work/List.jsx
```diff
@@ -1,53 +1,103 @@
...
+    return {
+      id: res._id,
+      title: res.title,
+      updatedAt: res.modified_date,
+      representativeImage: { id: res.representative_file_set_id, value: null },
+      manifestUrl: res.iiif_manifest,
+      published: res.published,
+      visibility: res.visibility.id
+        ? res.visibility
+        : { id: res.visibility, label: res.visibility },
```
adamjarling, line 22: @kdid @mbklein  Is there someplace we can see/track which properties will always come along for the ride on an Elasticsearch result data structure?   Via Kibana, or is there some schema definition file someplace that details this?

@katdivyareddy10 For example here, some of my results (could be just one), doesn't have `visibility`, so it's crashing.  Why don't we wrap the entire object preparation in a `try {} catch {}` to get around crashing, then individually guard against the fields being present?    

Here's we should check for `res.visibility` on line 20, then let the logic flow off that.

adamjarling: Also, this is my doing from the past, but could you make a guard against a `false` or "" value coming into this helpers file?  As long as we're in it...

image

Either give a default assignment in the parameter (`visibility = ""`, which might not work), or on the first line of the function:
```
if (!visibility) return;
```

adamjarling:
@katdivyareddy10 Let me try to wipe out my indexed data before acting on this.  Maybe it's something weird on my end with `visibility`.

adamjarling:
@katdivyareddy10 Ok, this is problem is verified after dropping my database and starting fresh.  Can you drop your db, start with some fresh data and trace the steps above to track down best way to handle?

I think the thing to keep in mind is we want to make the List/View cards as simple, yet robuts, as possible.

assets/js/screens/Work/List.jsx
kdid: We had to remove visibility briefly while we redo the backend. I'm not sure that there is a way to see the mappings via Kibana.

You could look at the code where we do the indexing to see the current state:
https://github.com/nulib/meadow/blob/deploy/staging/lib/meadow/data/schemas/work.ex#L82

We can try to be better about giving a heads up.

kdid: or we could index static values in the Elasticsearch index for now if it would help...
like "visibility": "NOT IMPLEMENTED"

adamjarling: @kdid Gotcha.  Actually the last option, just including the stubbed property of whatever datatype it's going to be, would be most helpful.  

So if it's a `string`, then yea ` { visibility: "WHATEVER" }` is fine.

If it's an object like in the mock data, then `{ visibility: { label: "", id: "" } }` would be most helpful.

We'd like to find a balance between defensive coding, and not creating a bunch of bloat that eventually won't be needed.

Hopefully this makes sense?   Lemme know any questions.  Thanks!

@katdivyareddy10  and myself are both looking at this now.

katdivyareddy10:
@adamjarling  Can you get the latest version of this branch and check if there's any problem? You would need to do a `down -v` to test this out as I pulled Karen's code.

assets/js/components/Work/UIWorkCardItem.jsx
```diff
@@ -1,64 +1,78 @@
...
             <img
               src={`${
-                work.representativeImage
-                  ? work.representativeImage + "/full/1280,960/0/default.jpg"
-                  : "/images/1280x960.png"
+                representativeImage.id
+                  ? iiifServerUrl +
+                    representativeImage.id +
+                    "/square/500,500/0/default.jpg"
+                  : representativeImage + "/full/1280,960/0/default.jpg"
```
adamjarling, line 32: Can we get rid of the IIIF Context usage now that we have `representativeImage.url`?

assets/js/components/Work/UIWorkCardItem.jsx
```diff
@@ -1,64 +1,78 @@
...
 import { Link } from "react-router-dom";
 import { setVisibilityClass, formatDate } from "../../services/helpers";
+import { IIIFProvider, IIIFContext } from "../../components/IIIF/IIIFProvider";
 
-const WorkCardItem = ({ work }) => {
+const WorkCardItem = ({
+  id,
+  representativeImage,
+  title,
+  descriptiveMetadata,
```
adamjarling, line 11: Do we need `descriptiveMetadata`?   Looks like it's just used for it's `title` property?  Wonder how it's different from `title` in this object?   I'd say let's lose the larger (in the future) `descriptiveMetadata`, and keep it thin.

assets/js/components/Work/UIWorkCardItem.jsx
```diff
@@ -1,64 +1,78 @@
...
+  workType,
+  visibility,
+  published,
+  accessionNumber,
+  fileSets,
```
adamjarling, line 16: We don't need all the `fileSets` and their contents, right?  This could get real heavy if there are 1000 or more `fileSets` per Work.  Let's just send in the total number of fileSets.

assets/js/components/Work/UIWorkCardItem.jsx
```diff
@@ -1,64 +1,78 @@
...
             />
           </Link>
         </figure>
       </div>
       <div className="card-content">
         <h3 className="title is-size-4">
-          {work.descriptiveMetadata.title
-            ? work.descriptiveMetadata.title
-            : "Untitled"}
+          {descriptiveMetadata.title ? descriptiveMetadata.title : "Untitled"}
```
adamjarling, line 42: Along the lines of above, let's figure out what `title` is the official title which should be used.  Maybe ping @kdid ?

assets/js/components/Work/UIWorkListItem.jsx
```diff
@@ -1,48 +1,62 @@
...
+  representativeImage,
+  title,
+  descriptiveMetadata,
+  workType,
+  visibility,
+  published,
+  accessionNumber,
+  fileSets,
+  manifestUrl,
+  updatedAt,
```
adamjarling, line 18: Same comments as the Card item above for thinning this out where possible.

assets/js/components/Work/UIWorkListItem.jsx
```diff
@@ -1,48 +1,62 @@
...
               <img
                 src={`${
-                  work.representativeImage
-                    ? work.representativeImage + "/full/1280,960/0/default.jpg"
-                    : "/images/1280x960.png"
+                  representativeImage.id
+                    ? iiifServerUrl +
+                      representativeImage.id +
+                      "/square/500,500/0/default.jpg"
+                    : representativeImage + "/full/1280,960/0/default.jpg"
```
adamjarling, line 36: Same as above comment regarding IIIF Context removal?

assets/js/components/Work/UIWorkListItem.jsx
```diff
@@ -1,48 +1,62 @@
...
-                  className={`tag ${setVisibilityClass(work.visibility.id)}`}
+                  className={`tag ${setVisibilityClass(visibility.id)}`}
                 >
-                  {work.visibility.label.toUpperCase()}
+                  {visibility.label.toUpperCase()}
                 </span>
               )}
-              {work.published && (
+              {published && (
                 <span data-testid="dd-published" className="tag is-success">
```
adamjarling, line 60: For the test ids, the `dd-` prefix could be confusing and might now be out of context.  Let's make it something more generic like `result-item-published`, `result-item-accession-number`, etc.?

assets/js/components/Work/UIWorkListItem.jsx
```diff
@@ -59,22 +73,18 @@ const WorkListItem = ({ work }) => {
               </thead>
               <tbody>
                 <tr>
-                  <td data-testid="dd-accession-number">
-                    {work.accessionNumber}
-                  </td>
+                  <td data-testid="dd-accession-number">{accessionNumber}</td>
```
adamjarling, line 76: Same here for all `data-testid`s...

assets/js/screens/Work/List.jsx
```diff
@@ -1,53 +1,101 @@
...
+      updatedAt: res.modified_date,
+      representativeImage: res.representative_file_set,
+      manifestUrl: res.iiif_manifest,
+      published: res.published,
+      visibility: res.visibility_term,
+      descriptiveMetadata: { title: res.title, description: res.description },
+      fileSets: res.file_sets,
+      accessionNumber: res.accession_number,
+      workType: res.work_type,
+    };
```
adamjarling, line 25: Let's make this match the code mentioned above.  We want to send in as few, lightweight props, as necessary to maintain performance.

adamjarling:
image

assets/js/components/Work/UIWorkCardItem.jsx
katdivyareddy10: `descriptiveMetadata` contains title and description fields of a work object. For search results, they are returned separately. For the actual work object in GraphiQL schema, `descriptiveMetadata` is an object with title, description and other fields.

assets/js/components/Work/UIWorkCardItem.jsx
katdivyareddy10: This is a good one, I'll make this change.

assets/js/components/Work/UIWorkCardItem.jsx
kdid: Yeah. `descriptiveMetadata` is not really a concept in the Elasticsearch documents

PR #780, opened 2020-05-07
Initial controlled vocabulary database modeling

bmquinn:
- Adds schemas and migrations for `CodedTerm`, `ControlledMetadataEntry`, `Field`, and `Value`.
- Adds `visibility` and `work_type` core metadata fields to the `Work` schema.
- Adds `preservation_level` and `status` to the `WorkAdministrativeMetadata` embedded schema.
- Adds `license` and `rights_statement` to the `WorkDescriptiveMetadata` embedded schema.
- 100% test coverage

Note: you may have to run `MIX_ENV=test mix ecto.reset` before running the tests locally.


lib/meadow/data/schemas/field.ex
```diff
@@ -0,0 +1,33 @@
...
+
+  import EctoEnum
+
+  defenum(MetadataClassEnum,
+    administrative: "administrative",
+    core: "core",
+    descriptive: "descriptive"
+  )
+
+  {@primary_key({:id, :string, autogenerate: false})}
```
mbklein, line 16: Why is this whole thing wrapped in `{}`? Do we do that elsewhere? Is it a best practice?

bmquinn: Must have been a copy pasta thing, fixed!

PR #822, opened 2020-05-20
Coded Term Functionality

mbklein:
* Seed `coded_terms` table from JSON files
* Implement `Meadow.Data.CodedTerms` context
* Implement GraphQL resolvers for `codeList` and `fetchCodedTermLabel` queries
* Remove `codeList` and `fetchCodedTermLabel` queries from the client mock
* Replace `mix run priv/repo/seeds.exs` with `mix meadow.seed`
* Call `meadow.seed` from `ecto.setup` (and therefore `meadow.setup`)
* Create a `Meadow.ReleaseTasks.seed/0` function to run seeds on a deployed instance


mix.exs
```diff
@@ -118,6 +118,7 @@ defmodule Meadow.MixProject do
         "meadow.ldap.setup test/fixtures/ldap_seed.ldif",
         "ecto.create --quiet",
         "ecto.migrate",
+        "run priv/repo/seeds.exs",
```
kdid, line 121: It's not running for me as part of `mix meadow.setup`. I had to run it manually. Is it supposed to be?

mbklein: Nope, it's definitely supposed to be part of setup. I'll add it in.

kdid:
Remove the `@client` or else the front end will break since the mock was removed:
https://github.com/nulib/meadow/blob/deploy/staging/assets/js/components/Work/controlledVocabulary.query.js#L21

PR #827, opened 2020-05-20
Seed fields table

mbklein:
* Extract log level wrapper from `coded_terms.exs` for reusability
* Improve `ReleaseTasks.seed()` handling of multiple seed names
* Seed fields using data from spreadsheet


mbklein:
> I dig that logging utility, too!

Thanks. The output from seeding was _super_ ugly and unreadable at debug level and I couldn't stand to leave it like that.

PR #828, opened 2020-05-20
Remove Jason.encode! from S3 body

mbklein:
The spec for `ExAws.S3.put_bucket_policy` is defined as
```elixir
put_bucket_policy(bucket :: binary, policy :: map())
```
so Dialyzer complains if `policy` is binary JSON, even though it actually _works_ OK (because `ExAws.request!` is happy to encode a map _or_ take an already-encoded binary representation).


mbklein:
Ugh. Forget it. `ExAws.S3.put_bucket_policy` wants a Map, but `ExAws.Auth.Utils.hash_sha256` wants a binary. We can't win.

PR #833, opened 2020-05-22
665 controlled vocab component

adamjarling:
This wires up "Controlled Term" metadata items into the Work form:

image

It uses a new wrapper component `<ControlledTermArray />` to handle the items.


assets/js/client-mocks.js
```diff
@@ -0,0 +1,1907 @@
+export const codeListMock = (scheme) => {
```
kdid, line 1: I didn't think this file (the mocks for the dropdowns) `assets/js/client-mocks.js` was needed anymore. Not sure if you are intentionally adding them back in?

adamjarling: @kdid I didn't intentionally add it (I don't think), but if it's not needed then yea I'll get rid of it.

PR #903, opened 2020-06-29
Update a missing downshift package causing an error; upgraded front-e…

adamjarling:
…nd dependencies; updated to OpenSeadragonViewer v2.1.0

This PR:
- Update a missing downshift package from the `package.json` file, causing an error in Controlled Term component; 
- upgraded front-end dependencies
- updated to OpenSeadragonViewer v2.1.0


katdivyareddy10:
It works as expected. Just curious about the title. If I add a work title, it looks like viewer doesn't get it even after I reload the work (gave it sometime before reload to make sure it gets there). 
Screen Shot 2020-06-30 at 9 20 43 AM

adamjarling:
@katdivyareddy10 @davidschober @mbklein Ah, so the Viewer is pulling in it's title from the root `label` property of a manifest.  So this update would depend on how manifests are updated when data changes.   We can work around it locally, but good to know how manifests are being updated when we write stuff against it moving forward.

PR #916, opened 2020-07-02
Replace the ConCache package with Cachex

mbklein:
Completes repodev_planning_and_docs#882


kdid:
Can you please write up an issue for this and connect it? I wrote up an issue for the ETS backed coded term cache but I think that encompasses more than just this.

PR #936, opened 2020-07-09
End to end coded term implementation.

kdid:
Editing and saving works with coded terms now works from UI
 - rights statement
 - license
 - visibility
 - work type (saving only not editable)
 - preservation level 
 - status

Updated Elasticsearch indexing structure to more closely resemble GraphQL structure

**_Database changes so staging data will need to be reset on deploy and dev environments will require a `devstack down -v`_**

Changes: 
- Updated API and front end components.
- Indexing coded terms for file sets and works.
- Removes "Not Live" and "Mock" on properties and NOT IMPLEMENTED documentation and descriptions
- Updated Ecto changesets for works and file sets

We broke out a separate issue to remove Apollo client mocking artifacts, the tests seemed to have a dependency on them and we weren't sure why: https://github.com/nulib/repodev_planning_and_docs/issues/898

### DIFF
--- a/assets/js/components/UI/CodedTerm/Item.jsx
+++ b/assets/js/components/UI/CodedTerm/Item.jsx
@@ -2,6 +2,8 @@ import React from "react";
 import PropTypes from "prop-types";
 
 const UICodedTermItem = ({ item = {} }) => {
+  if (!item) return null;
+
   if (!item.id) return null;
 
   return <p>{item.label}</p>;
--- a/assets/js/components/Work/Tabs/About/CoreMetadata.jsx
+++ b/assets/js/components/Work/Tabs/About/CoreMetadata.jsx
@@ -23,6 +23,13 @@ const WorkTabsAboutCoreMetadata = ({
   } = useQuery(CODE_LIST_QUERY, {
     variables: { scheme: "RIGHTS_STATEMENT" },
   });
+  const {
+    loading: licenseLoading,
+    error: licenseError,
+    data: licenseData,
+  } = useQuery(CODE_LIST_QUERY, {
+    variables: { scheme: "LICENSE" },
+  });
 
   return showCoreMetadata ? (
     <div className="columns is-multiline">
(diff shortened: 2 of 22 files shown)


assets/js/components/UI/CodedTerm/Item.jsx
```diff
@@ -2,6 +2,8 @@ import React from "react";
 import PropTypes from "prop-types";
 
 const UICodedTermItem = ({ item = {} }) => {
+  if (!item) return null;
```
adamjarling, line 5: Not that it matters, but this condition will never return `true` with the `item = {}` default declaration above.  Probably could remove it from line 4 to read `({ item })` instead.

assets/js/components/Work/Tabs/About/CoreMetadata.jsx
```diff
@@ -104,9 +115,20 @@ const WorkTabsAboutCoreMetadata = ({
...
+              name="license"
+              label="License"
+              options={licenseData ? licenseData.codeList : []}
+              defaultValue={
+                descriptiveMetadata.licenseStatement
+                  ? descriptiveMetadata.licenseStatement.id
+                  : ""
+              }
+              errors={errors}
+            />
```
adamjarling, line 131: I think in standup you mentioned a "helper" default option for `select` form elements.  This is the `prop` which will give an empty initial value:

https://github.com/nulib/meadow/blob/81fee1555612b51228116a5c6ea69d866f830322/assets/js/components/UI/Form/Select.jsx#L14

assets/js/components/UI/CodedTerm/Item.jsx
kdid: Hmm. We added it because we we're getting an error.

adamjarling: Weird.  An empty object literal shouldn't return `undefined` or `null` or be falsy.

adamjarling: ...unless maybe the item being passed in is `null`, then maybe it'd bypass the default value?  Probably not that big a deal.

PR #954, opened 2020-07-15
SetCurrentUser plug lookup user from cache first

kdid:
- Store user/role information in an ETS cache with an ttl of 20 minutes in order to prevent LDAP timeouts that are occurring with user lookup on every GraphQL request.

### DIFF
--- a/lib/meadow_web/plugs/set_current_user.ex
+++ b/lib/meadow_web/plugs/set_current_user.ex
@@ -29,6 +29,18 @@ defmodule MeadowWeb.Plugs.SetCurrentUser do
   end
 
   defp refresh_user(%{username: username}) do
+    case Cachex.get!(Meadow.Cache.Users, username) do
+      nil ->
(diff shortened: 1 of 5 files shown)


lib/meadow_web/plugs/set_current_user.ex
```diff
@@ -29,6 +29,18 @@ defmodule MeadowWeb.Plugs.SetCurrentUser do
   end
 
   defp refresh_user(%{username: username}) do
+    case Cachex.get!(Meadow.Cache.Users, username) do
```
mbklein, line 32: There's also [`Cachex.fetch`](https://hexdocs.pm/cachex/Cachex.html#fetch/4), which uses the "get if it's there; generate and put if it's not" pattern. I think the main difference between that and what you've done is that yours renews the 20 minute timeout on every request, which might be better. This isn't a requested/suggested change so much as a discussion of `Cachex` since we seem to be finding new places to use it.

PR #955, opened 2020-07-15
Local cache setup for schemes in descriptive metadata

katdivyareddy10:
- [x] Set up a local caching approach which checks if the Marc records are cached locally
- [x]  Use local cached values if they exist
- [x]  Call the GraphQL query if they don't exist, and put in local cache
- [x]  Provide some indication in the UI the user is using a cached version of Marc records (which maybe they can view in a pop up modal)?

### DIFF
--- a/assets/js/components/Work/Tabs/About/DescriptiveMetadata.jsx
+++ b/assets/js/components/Work/Tabs/About/DescriptiveMetadata.jsx
@@ -6,7 +6,7 @@ import UIControlledTermList from "../../../UI/ControlledTerm/List";
 import { CODE_LIST_QUERY } from "../../controlledVocabulary.gql.js";
 import UIFormFieldArrayDisplay from "../../../UI/Form/FieldArrayDisplay";
 import UIFormControlledTermArray from "../../../UI/Form/ControlledTermArray";
-import { useQuery } from "@apollo/react-hooks";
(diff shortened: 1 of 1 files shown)


assets/js/components/Work/Tabs/About/DescriptiveMetadata.jsx
```diff
@@ -19,29 +28,95 @@ const WorkTabsAboutDescriptiveMetadata = ({
...
-    data: authorityData,
-    loading: authorityLoading,
-    errors: authorityErrors,
-  } = useQuery(CODE_LIST_QUERY, { variables: { scheme: "AUTHORITY" } });
+  const localList = localStorage.getItem("codeLists");
+  const [codeLists, setCodeLists] = useState(
+    localList ? JSON.parse(localList) : {}
+  );
+  useEffect(() => {
+    localStorage.setItem("codeLists", JSON.stringify(codeLists));
```
adamjarling, line 36: This is going to set `localStorage.setItem` every single time the component loads.  We don't really want to re-set the `localStorage` value if it already exists, right?   Let's find a way around this.

adamjarling: Maybe before calling line 36, we check 1.) `codeLists` doesn't exist or have values, or 2.) the user clicked the "refresh cache values" button?

PR #972, opened 2020-07-20
Create mix tasks to make it easy to seed Meadow with data from Digital Collections

mbklein:
## Description

- Create, validate, and approve ingest sheet
- Add progress to seed_data; wait for ingest to complete
- Use the CSV name as the ingest sheet name
- Create fixtures for seed data test
- Fix seed_data test sandboxing/cleanup and exclude it from running by default
- Create `mix.seed_data` and `mix.dc_download` tasks

### Create some seed data

- Run `mix meadow.dc_download wwii_posters faf4f60e-78e0-4fbf-96ce-4ca8b4df597a 10`
- Full usage for the mix task can be displayed by running `mix help meadow.dc_download`

### Import seed data into Meadow

1. Make sure devstack is up and initialized
2. `mix meadow.seed_data wwii_posters`
3. Wait for import to finish, start the server, and go check out the results

### Notes

- Descriptive and Administrative metadata are not brought over. Only images, groupings of filesets into works, work accession numbers, and original filenames.
- If `phx.server` is running while you run `mix meadow.seed_data`, you can watch the seed data ingest progress in the UI just like any other ingest sheet
- If there are validation errors (e.g., trying to ingest the same data twice or accession number collisions), they will be written to the console and the import will fail


kdid:
I think this is really great!

- Does it make sense to have the path to the seed data directory be inside the meadow project and gitignore it? Just thinking it might make the command easier for all devs. 

- Please document in README? Also in docs site? (at least a callout saying you can add seed data to the dev env and a link to the readme for instructions?)

- I accidentally used a staging pid and it created an (empty) ingest sheet and then began the upload and just said 0% forever. If it's easy maybe it could return a message that it failed to retrieve any data or that the sheet was empty. But if not forget it....

mbklein:
Great comments.

* I'll add `priv/seed_data` to `.gitignore`. That way, if anyone wants to have the seed data there in the project, there's a good place for it. `mix meadow.dc_download priv/seed_data/whatever ...` will already put data there.

* Thanks for the documentation reminder. The tasks themselves have pretty comprehensive `@moduledoc`s – do you think a note to run `mix help meadow.dc_download` and `mix help meadow.seed_data` would be sufficient?

* Yep, I'll add an “empty results” check right now. @bmquinn and I discussed that yesterday but forgot to do it.

mbklein:
> * I'll add `priv/seed_data` to `.gitignore`. That way, if anyone wants to have the seed data there in the project, there's a good place for it. `mix meadow.dc_download priv/seed_data/whatever ...` will already put data there.

Or do you think we should _always_ use priv/seed_data as the base directory just to make things _easy_ easy?

PR #976, opened 2020-07-21
Wire up beginnings of Reactivesearch faceting into the app

adamjarling:
Puts some general structure in place for implementing multiple Reactive List "Multilist" components.

image


kdid:
Approving. Looks good. I would say show the label instead of the id for visibility but this is probably just your first step.

PR #978, opened 2020-07-22
899 remove firstitem select on dropdown

katdivyareddy10:
Screen Shot 2020-07-22 at 8 48 53 AM
Screen Shot 2020-07-22 at 8 48 42 AM


kdid:
@katdivyareddy10 - These are not required fields (no fields are actually required until the work is set to "published". I'm wondering if you made them required just so the form wouldn't throw an error if it was submitted with blanks?

PR #980, opened 2020-07-22
Add local controlled terms wireframes to sketch doc

bmquinn:
- Add local controlled terms wireframes to sketch doc
- Add image exports for the new sketches in `doc/wireframes/exports/controlled_vocabulary/`
- Updates gitignore to exclude sketch exports


davidschober:
These look pretty good. I wonder if we should leave the PR open until after the meeting tomorrow in case we need to clean up a bit.

PR #982, opened 2020-07-23
Validate coded term input through GraphQL

kdid:
- Validate coded term input through the GraphQL API. 
- Provide helpful error messages if input fails validation
- Add Elixir tests for `CodedTerm` type
- Removed `required` from form input properties
- Updates to front end coded term UI to support API changes
- Removes extra s from 'In Progress' in Status seed data

Closes https://github.com/nulib/repodev_planning_and_docs/issues/935

### DIFF
--- a/assets/js/components/Work/Tabs/About.jsx
+++ b/assets/js/components/Work/Tabs/About.jsx
@@ -148,10 +148,12 @@ const WorkTabsAbout = ({ work }) => {
         identifier,
         keywords,
         legacyIdentifier,
-        license: {
-          id: data.license,
-          scheme: "LICENSE",
-        },
+        license: data.license
+          ? {
+              id: data.license,
+              scheme: "LICENSE",
+            }
+          : {},
         notes,
         physicalDescriptionMaterial,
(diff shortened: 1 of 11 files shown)


assets/js/components/Work/Tabs/About.jsx
```diff
@@ -160,10 +162,12 @@ const WorkTabsAbout = ({ work }) => {
...
-        rightsStatement: {
-          id: data.rightsStatement,
-          scheme: "RIGHTS_STATEMENT",
-        },
+        rightsStatement: data.rightsStatement
+          ? {
+              id: data.rightsStatement,
+              scheme: "RIGHTS_STATEMENT",
+            }
+          : {},
```
adamjarling, line 170: So it's easier for the API/backend to handle an empty object as opposed to `null`?  If so, works for us.

kdid: Well, it could do either. I wasn't sure which was the best or most consistent honestly. But I got two votes for doing it the empty object way yesterday in Slack...

PR #993, opened 2020-07-28
Update indexing to support front end faceting needs

kdid:
Indexes `display_facet` for controlled fields to give Fen (and potentially internal Meadow) the role concatenated with the label to use/display in facets.

Screen Shot 2020-07-28 at 10 46 48 AM

### DIFF
--- a/lib/meadow/data/schemas/work_descriptive_metadata.ex
+++ b/lib/meadow/data/schemas/work_descriptive_metadata.ex
@@ -118,7 +118,15 @@ defmodule Meadow.Data.Schemas.WorkDescriptiveMetadata do
     def encode_field([field | []]), do: [encode_field(field)]
     def encode_field([field | fields]), do: [encode_field(field) | encode_field(fields)]
 
-    def encode_field(%ControlledMetadataEntry{} = field), do: Map.from_struct(field)
+    def encode_field(%ControlledMetadataEntry{role: nil} = field) do
+      Map.from_struct(field)
+      |> Map.put(:displayFacet, field.term.label)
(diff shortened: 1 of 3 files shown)


lib/meadow/data/schemas/work_descriptive_metadata.ex
```diff
@@ -118,7 +118,15 @@ defmodule Meadow.Data.Schemas.WorkDescriptiveMetadata do
     def encode_field([field | []]), do: [encode_field(field)]
     def encode_field([field | fields]), do: [encode_field(field) | encode_field(fields)]
 
-    def encode_field(%ControlledMetadataEntry{} = field), do: Map.from_struct(field)
+    def encode_field(%ControlledMetadataEntry{role: nil} = field) do
+      Map.from_struct(field)
+      |> Map.put(:display_facet, field.term.label)
```
bmquinn, line 123: Should this be camelCased as `displayFacet` to match the other fields?

kdid: argh! yes. Thanks!

bmquinn: No problem! Not sure if that'll affect the `*facet` change in `meadow.json` or not, but you may want to verify.

kdid: arg. yes it will. thanks.

adamjarling:
@kdid Want to get this merged before we start adding additional Facet items to the Meadow search page?

PR #1000, opened 2020-07-29
Related URL and label backend functionality

kdid:
- Backend for storing `related_url` property as uri/label pair (with label as a controlled term)
 - updates the Meadow seed to add related urls to the coded terms table
 - adds related url to work object in GraphQL api
 - removes front end code for plain text related url field.

Screen Shot 2020-07-29 at 8 08 33 AM

**Reviewers and after merge** - you'll need to do a `devstack down -v` or (otherwise reset your database and index)

### DIFF
--- a/priv/elasticsearch/meadow.json
+++ b/priv/elasticsearch/meadow.json
@@ -137,6 +137,7 @@
         {
           "labels": {
             "path_match": "*.label",
+            "path_unmatch": "descriptiveMetadata.relatedUrl.label",
             "mapping": {
               "type": "text",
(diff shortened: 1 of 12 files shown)


priv/elasticsearch/meadow.json
```diff
@@ -137,6 +137,7 @@
         {
           "labels": {
             "path_match": "*.label",
+            "path_unmatch": "descriptiveMetadata.relatedUrl.label",
```
mbklein, line 140: 👏

PR #1025, opened 2020-08-03
Add components and tests for batchedit and work tabs about page

katdivyareddy10:
New components and tests added as per comments in https://northwestern.app.box.com/notes/691882895514?s=kqfgupsbacut195skpyowxq5bu38u83v

Screen Shot 2020-08-03 at 8 54 09 AM
Screen Shot 2020-08-03 at 8 54 30 AM
Screen Shot 2020-08-03 at 8 55 50 AM

### DIFF
--- a/assets/js/components/BatchEdit/About/About.jsx
+++ b/assets/js/components/BatchEdit/About/About.jsx
@@ -1,32 +1,18 @@
-import React, { useState, useEffect } from "react";
+import React from "react";
 import PropTypes from "prop-types";
 import { toastWrapper } from "../../../services/helpers";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { useForm } from "react-hook-form";
-import { useMutation } from "@apollo/client";
 import UITabsStickyHeader from "../../UI/Tabs/StickyHeader";
 import BatchEditAboutCoreMetadata from "./CoreMetadata";
-import BatchEditDescriptiveMetadata from "./DescriptiveMetadata";
-import UIError from "../../UI/Error";
-
-/** @jsx jsx */
-import { css, jsx } from "@emotion/core";
+import BatchEditAboutControlledMetadata from "./ControlledMetadata";
+import BatchEditAboutUncontrolledMetadata from "./UncontrolledMetadata";
+import BatchEditAboutPhysicalMetadata from "./PhysicalMetadata";
+import BatchEditAboutRightsMetadata from "./RightsMetadata";
+import BatchEditAboutIdentifiersMetadata from "./IdentifiersMetadata";
+import UIAccordion from "../../UI/Accordion";
 
 const BatchEditAbout = ({ numberOfResults }) => {
-  // Whether box dropdowns are open or closed
-  const [showCoreMetadata, setShowCoreMetadata] = useState(true);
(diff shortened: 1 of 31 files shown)


assets/js/components/BatchEdit/About/About.jsx
```diff
@@ -92,43 +78,76 @@ const BatchEditAbout = ({ numberOfResults }) => {
         </span>
         You are editing {numberOfResults} items. Proceed with caution.
       </p>
+      <UIAccordion
```
adamjarling, line 81: This `UIAccordion`s are great.  Curious when would the `isVisible` prop be `false`?

assets/js/components/BatchEdit/About/CoreMetadata.jsx
```diff
@@ -43,59 +42,57 @@ const BatchEditAboutCoreMetadata = ({ errors, register, ...restProps }) => {
...
+          name="alternateTitle"
+          data-testid="alternate-title"
+          label="Alternate Title"
+          errors={errors}
+        />
       </div>
       <div className="column is-half">
         {/* Date Created */}
-        <UIFormField label="Date Created" notLive>
+        <UIFormField label="Date Created" required notLive>
```
adamjarling, line 59: Does this need to be required?

assets/js/services/metadata.js
```diff
@@ -1,3 +1,82 @@
+export const CONTROLLED_METADATA = [
```
adamjarling, line 1: Looks like `CONTROLLED_METADATA` and `DESCRIPTIVE_METADATA.controlledTerms` are tracking the same thing and both being used?  This may lead to confusion moving forward.   Can we consolidate into using one or the other?

assets/js/components/BatchEdit/About/About.jsx
katdivyareddy10: `isVisible` is always true for the way we're using it but I thought if we have to use it so that the accordions load as closed by default, we could use the prop.

assets/js/components/BatchEdit/About/CoreMetadata.jsx
katdivyareddy10: I think so from the docs..
Screen Shot 2020-08-03 at 12 54 27 PM

assets/js/services/metadata.js
katdivyareddy10: So the idea was to get rid of `DESCRIPTIVE_METADATA` as a whole when the tests in localstorage are fixed. Looks like I missed removing it from About component.New commit up in a few.

assets/js/components/BatchEdit/About/About.jsx
adamjarling: Ah, now I see what you mean.  To simplify this, what do you think about changing the `isVisible` prop (to me that sounds like the whole element, not necessarily the contents of the accordion component) to `defaultOpen`, which is a prop that defaults to `true`.   If a user wanted to render the component initially closed, then they could include the prop `defaultOpen={false}`?

assets/js/components/BatchEdit/About/CoreMetadata.jsx
kdid: Nothing is required until the point when the work is published. 

Date created is not yet implemented on the backend. (it's an EDTF date)

PR #1057, opened 2020-08-11
Add relatedUrl component, update tests and mocks

katdivyareddy10:
Screen Shot 2020-08-11 at 1 34 13 PM


kdid:
It looks great, did you want to validate that the input is a url? (Or maybe that's out of scope at this point so I'm approving anyways)

katdivyareddy10:
> It looks great, did you want to validate that the input is a url? (Or maybe that's out of scope at this point so I'm approving anyways)

Not sure how react-hook-form handles validation . I'd say, it might be out of scope for this issue.

PR #1070, opened 2020-08-13
Facet filter interaction

kdid:
- Get the facet counts to interact with each other. ("OR within a single filter (e.g. contributor) and AND with multiple filters")
- I tried to get the search bar to inteact with the facets as well but anytime I tried to add "SearchSensor" to the filter list I got a warning and it didn't work. So not sure what's diffrent between this and DC. (it works there)


adamjarling:
@kdid Does this seem to get the results you want?

PR #1077, opened 2020-08-17
Clean up grid item

katdivyareddy10:
Screen Shot 2020-08-17 at 10 25 15 AM

### DIFF
--- a/assets/js/components/Work/CardItem.jsx
+++ b/assets/js/components/Work/CardItem.jsx
@@ -10,15 +10,12 @@ const WorkCardItem = ({
   workType,
   visibility,
   published,
-  accessionNumber,
-  fileSets,
-  manifestUrl,
-  updatedAt,
+  collectionName,
 }) => {
   return (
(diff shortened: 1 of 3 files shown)


assets/js/components/Work/CardItem.jsx
```diff
@@ -33,11 +30,13 @@ const WorkCardItem = ({
...
-          className="title is-size-4"
-          dangerouslySetInnerHTML={{ __html: title ? title : "Untitled" }}
-        ></h3>
-
+        <h2
+          className="subtitle"
+          dangerouslySetInnerHTML={{
+            __html: title ? title : "Untitled",
+          }}
+        ></h2>
```
adamjarling, line 38: @katdivyareddy10 Just curious why we're doing the `dangerouslySetInnerHTML` here?

katdivyareddy10: This is existing code.

adamjarling: Got it, never mind:)

PR #1078, opened 2020-08-17
Adds and Deletes posting successfully to batch edit process

adamjarling:
This finalizes the process of preparing and submitting Batch Edit add and delete items to the GraphQL API.  

I tried to simplify some of contents of `assets/js/components/BatchEdit/About/Confirmation.jsx`, and how we're prepping data in `assets/js/components/BatchEdit/About/About.jsx` as well. 

image

### DIFF
--- a/assets/js/components/BatchEdit/About/Confirmation.jsx
+++ b/assets/js/components/BatchEdit/About/Confirmation.jsx
@@ -8,6 +8,18 @@ import { useMutation } from "@apollo/client";
 import BatchEditConfirmationTable from "./ConfirmationTable";
 import { removeLabelsFromBatchEditPostData } from "../../../services/metadata";
 
+/** @jsx jsx */
+import { css, jsx } from "@emotion/core";
+const headerWrapper = css`
+  display: flex;
+  align-items: center;
+  margin-bottom: 1rem;
+
+  span {
+    margin-left: 0.8rem;
+  }
+`;
+
 const BatchEditConfirmation = ({
   batchAdds,
   batchDeletes,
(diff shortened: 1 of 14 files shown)


kdid:
@adamjarling - In the add and remove screens in your screenshots above, I'd show users the Role Label rather than the Role Id. (There won't be any ambiguity regarding roles so they won't need to see the id to tell them apart).

adamjarling:
@kdid @davidschober @laura-alagna Any value in displaying the Role Scheme in the above screenshot?

davidschober:
I don’t think so.

kdid:
Let's see if we can get it working if we add a `query` to the query. Ex: 

` "{\"query\":{\"term\":{\"workType.id\": \"IMAGE\"}}}"`

adamjarling:
Yea, it needs to be set up like it currently is for the front-end to work, but I'll wrap the final POST to the mutation in a `query: { ...}`.

assets/js/components/BatchEdit/About/Confirmation.jsx
```diff
@@ -107,105 +90,61 @@ const BatchEditConfirmation = ({
...
-                      </li>
-                    )
-                )}
-            </ul>
-          </section> */}
+          {hasAdds && (
+            <section className="content">
+              <div css={headerWrapper}>
+                <FontAwesomeIcon icon="plus" size="2x" />
+                <span class="subtitle">Adding</span>
```
katdivyareddy10, line 97: can you change this to className?

assets/js/components/BatchEdit/About/Confirmation.jsx
```diff
@@ -107,105 +90,61 @@ const BatchEditConfirmation = ({
...
+                );
+              })}
+            </section>
+          )}
+
+          {hasDeletes && (
+            <section className={`content ${hasAdds ? "py-6" : ""}`}>
+              <div css={headerWrapper}>
+                <FontAwesomeIcon icon="minus-square" size="2x" />
+                <span class="subtitle">Removing</span>
```
katdivyareddy10, line 121: Here as well

assets/js/components/BatchEdit/About/Confirmation.jsx
adamjarling: Will do

kdid:
Functionality works now from the UI. Looks good.

PR #1080, opened 2020-08-17
Implement Meadow backend part of infrastructure for creating and indexing shareable links

kdid:
- Adds `createSharedLink` mutation, which will return a `sharedLinkId` which you can use to send a request to the proxy for a work.
- Expiration date defaults to 14 days (In future iterations, we can make the default overridable in the GraphQL API)

Example use:
```
mutation{
  createSharedLink(workId: "f92ebef9-0841-4120-a4ed-7a4a118acc50"){
    workId
    expires
    sharedLinkId
  }
}

# result
{
  "data": {
    "createSharedLink": {
      "expires": "2020-08-31T16:52:57.950026Z",
      "sharedLinkId": "e6ef76dc-021d-42cd-b2b7-a82f0736e460",
      "workId": "f92ebef9-0841-4120-a4ed-7a4a118acc50"
    }
  }
}
```

Also, indexes the shared link in a `shared_links` index. (The front end does not use the index)

Screen Shot 2020-08-17 at 12 03 20 PM


kdid:
@adamjarling - I don't see it on our Demo ticket and it looks like the front end part of the work was not pulled into current sprint so I don't think so unless @davidschober wants to demo the backend only. (which might be hard to understand for folks)

davidschober:
I wasn't sure we'd be able to get it all done. Let's keep running with what we have. If we have capacity, we can add it. But There's still a bit to work on from the frontend side.

PR #1081, opened 2020-08-17
Add default case for 0 filesets

katdivyareddy10:
Screen Shot 2020-08-17 at 3 10 51 PM

### DIFF
--- a/assets/js/components/IngestSheet/ApprovedInProgress.jsx
+++ b/assets/js/components/IngestSheet/ApprovedInProgress.jsx
@@ -28,6 +28,7 @@ const IngestSheetApprovedInProgress = ({ ingestSheet }) => {
     <UIProgressBar
       percentComplete={Number(ingestProgress.percentComplete)}
       totalValue={ingestProgress.totalFileSets}
+      isIngest={true}
     />
   );
 };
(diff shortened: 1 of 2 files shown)


kdid:
@Divya Katpally - on your PR, I'm wondering if we could make it say "Validating file sets" rather than "Ingesting file sets"? This step is pre-ingest. 

(Or maybe just leave that off altogether and just show percentage...)

katdivyareddy10:
> @divya Katpally - on your PR, I'm wondering if we could make it say "Validating file sets" rather than "Ingesting file sets"? This step is pre-ingest.
> 
> (Or maybe just leave that off altogether and just show percentage...)

@adamjarling I like the "Validating file sets" idea, what are your thoughts?

adamjarling:
@katdivyareddy10 Yea, the wording above sounds good to me.

adamjarling:
@katdivyareddy10 Does it seem like we're straying away from making this component generic?  Seems pretty tightly coupled to implementation (it was from the start, now seemingly more so).  Maybe in the future we pass in the message as a prop or something.

katdivyareddy10:
I agree, I tried to do that but realized that this is used only in 2 places and specifically for ingest sheet progress only.

PR #1085, opened 2020-08-18
Clean up batchedit remove and faceting

katdivyareddy10:
Screen Shot 2020-08-18 at 12 13 21 PM
Screen Shot 2020-08-18 at 12 13 31 PM

### DIFF
--- a/assets/js/components/BatchEdit/Remove.js
+++ b/assets/js/components/BatchEdit/Remove.js
@@ -32,11 +32,7 @@ export default function BatchEditRemove({
             {removeItems.map((item) => {
               const { label, role, term } = splitFacetKey(item);
 
-              return (
-                <li key={item} className="has-text-danger">
-                  {`${label} - ${term} - ${role}`}
-                </li>
-              );
+              return <li key={item}>{`${label} - ${term}`}</li>;
             })}
           </ul>
         </div>
--- a/assets/js/services/reactive-search.js
+++ b/assets/js/services/reactive-search.js
@@ -28,8 +28,8 @@ export const REACTIVE_SEARCH_THEME = {
 // Documention: https://docs.appbase.io/docs/reactivesearch/v3/theming/classnameinjection/
 const facetClasses = {
   checkbox: "facet-checkbox",
-  label: "facet-item-label",
-  title: "facet-title",
+  label: "facet-item-label mx-2",
+  title: "facet-title mt-4",
 };
 
 const defaultListItemValues = {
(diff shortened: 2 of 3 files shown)


assets/js/components/BatchEdit/Remove.js
```diff
@@ -27,15 +27,15 @@ export default function BatchEditRemove({
...
-        <div className="content">
-          <ul>
+        <div className="content has-background-danger-light">
+          <ul className="py-4">
             {removeItems.map((item) => {
               const { label, role, term } = splitFacetKey(item);
 
               return (
-                <li key={item} className="has-text-danger">
-                  {`${label} - ${term} - ${role}`}
```
adamjarling, line 37: @katdivyareddy10 @kdid  Do we want to keep `role` here for Contributor and Subject s?

assets/js/services/reactive-search.js
```diff
@@ -28,8 +28,8 @@ export const REACTIVE_SEARCH_THEME = {
 // Documention: https://docs.appbase.io/docs/reactivesearch/v3/theming/classnameinjection/
 const facetClasses = {
   checkbox: "facet-checkbox",
-  label: "facet-item-label",
-  title: "facet-title",
+  label: "facet-item-label mx-2",
+  title: "facet-title mt-4",
```
adamjarling, line 32: Clever!   Nice one.

assets/js/components/BatchEdit/Remove.js
katdivyareddy10: I can remove it if it's not useful.

kdid: Keep the roles for contributor and subject. (only display the role label (not role id)) but removing:
Contributor 1 Term - contributor1id - (Author) 
Contributor 1- contributor1id -  (Publisher)

are two different things - users need to see the difference.

katdivyareddy10: Ok, I guess that means we're keeping role.

katdivyareddy10: If there are no other changes, could either of you review this PR?

assets/js/components/BatchEdit/Remove.js
```diff
@@ -27,15 +27,15 @@ export default function BatchEditRemove({
...
-                </li>
+                <li key={item}>{`${label} - ${term} ${
+                  role ? "- " : ""
+                } ${role}`}</li>
```
adamjarling, line 38: @katdivyareddy10 It's small, but can we tighten up this ternary?  To look something like:
```
<li key={item}>{`${label} - ${term}`} {role ? `- ${role}` : ""}</li>
```

PR #1093, opened 2020-08-19
Add shareable link data to Work screen

adamjarling:
Wired up to the Work screen.

image

### DIFF
--- a/assets/js/components/Work/SharedLinkNotification.jsx
+++ b/assets/js/components/Work/SharedLinkNotification.jsx
@@ -1,0 +1,30 @@
+import React from "react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import PropTypes from "prop-types";
+import moment from "moment";
+
+export default function WorkSharedLinkNotification({ linkData }) {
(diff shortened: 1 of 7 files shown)


assets/js/components/Work/SharedLinkNotification.jsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import PropTypes from "prop-types";
+import moment from "moment";
+
+export default function WorkSharedLinkNotification({ linkData }) {
+  const linkUrl = `http://meadow.rdc-staging.library.northwestern.edu/shared/${linkData.sharedLinkId}`;
```
kdid, line 7: Did you want to change it to `http://digital-collections.rdc-staging.library.northwestern.edu/shared/`

(I realize it won't work yet and the value should be coming from an env variable or GQL query but do you think it would help for the demo to say "This will be a link to the work in the Digital Collections site?" )

adamjarling: Ah, good call

PR #1135, opened 2020-09-02
Speed up ingest validation by reducing the number of AWS requests

bmquinn:
Make the validation more efficient with a single `ExAws.S3.list_objects/1` request, instead of an `ExAws.S3.head_object/1` request for each file in the ingest sheet.


mbklein (approved):
Sweet!

PR #1153, opened 2020-09-09
Ingest speed and reliability improvements

mbklein:
* Add a `retry` status to the pipeline
* Use timeout / retry to avoid stream lockups in `GenerateFileSetDigests`


bmquinn (approved):
Looks great!
checksum

PR #1192, opened 2020-09-16
Connect search box to facet items

adamjarling:
This behavior now seems to work as Digital Collections does with filtering the combination of Search terms and Facet selections.

image

image

### DIFF
--- a/assets/js/components/Search/FacetSidebar.jsx
+++ b/assets/js/components/Search/FacetSidebar.jsx
@@ -1,11 +1,16 @@
 import React from "react";
 import { MultiList } from "@appbaseio/reactivesearch";
-import { FACET_SENSORS } from "../../services/reactive-search";
+import { FACET_SENSORS, SEARCH_SENSOR } from "../../services/reactive-search";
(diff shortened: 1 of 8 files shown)


adamjarling:
Quick question @davidschober ... @katdivyareddy10 brought this up in a PR.   The question is, do we want the Search bar to be performing an "AND" or "OR" method?

Here's with "OR":

image

And here's with "AND":

image

And here's with "OR" again, with quotes, which I thought would try to search on the phrase, right?

image

davidschober:
ok. Something is odd there. Quotes should perform a phrase search. We need to figure that out.  I am leaning toward "and" qualifiers, but we would need to switch that in DC as well. For now let's keep "OR" as the default qualifier but we need to make sure phrasal searching is working.

adamjarling:
Actually, never mind on the phrase search not working.  There's a config prop on the `DataSearch` component for `searchOperators` which defaults to `false`.  Just switched it on and works.

kdid (approved):

PR #1196, opened 2020-09-17
Add validation rules for URL with custom messages

katdivyareddy10:
Tested regex on https://regex101.com/

Screen Shot 2020-09-17 at 1 03 07 PM
Screen Shot 2020-09-17 at 1 03 47 PM

### DIFF
--- a/assets/js/services/helpers.js
+++ b/assets/js/services/helpers.js
@@ -23,8 +23,7 @@ export function formatSimpleISODate(date) {
 }
 
 export function isUrlValid(url) {
-  const res = url.match(URL_PATTERN_MATCH);
-  return res ? true : false;
+  return url.match(URL_PATTERN_MATCH) ? true : false;
 }
 
 export function getClassFromIngestSheetStatus(status) {
--- a/assets/js/services/helpers.js
+++ b/assets/js/services/helpers.js
@@ -1,5 +1,6 @@
 import moment from "moment";
 import { toast } from "bulma-toast";
+import { URL_PATTERN_MATCH } from "./global-vars";
 
 /**
  * Escape double quotes (which may interfere with Search queries)
@@ -18,8 +19,12 @@ export function formatDate(date) {
 export function formatSimpleISODate(date) {
   if (!date) return "";
   let newDate = moment(date).format().substring(0, 16);
-
   return newDate;
+}
+
+export function isUrlValid(url) {
+  const res = url.match(URL_PATTERN_MATCH);
+  return res ? true : false;
 }
 
 export function getClassFromIngestSheetStatus(status) {
--- a/assets/js/services/helpers.test.js
+++ b/assets/js/services/helpers.test.js
@@ -3,12 +3,24 @@ import {
   escapeDoubleQuotes,
   formatDate,
   getClassFromIngestSheetStatus,
+  isUrlValid,
 } from "./helpers";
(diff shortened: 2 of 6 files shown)


assets/js/services/helpers.js
```diff
@@ -1,5 +1,6 @@
 import moment from "moment";
 import { toast } from "bulma-toast";
+import { URL_PATTERN_MATCH } from "./global-vars";
```
adamjarling, line 3: Nice.  We may need to re-use this...good call.

assets/js/services/helpers.js
```diff
@@ -18,10 +19,14 @@ export function formatDate(date) {
 export function formatSimpleISODate(date) {
   if (!date) return "";
   let newDate = moment(date).format().substring(0, 16);
-
   return newDate;
 }
 
+export function isUrlValid(url) {
+  const res = url.match(URL_PATTERN_MATCH);
+  return res ? true : false;
```
adamjarling, line 27: If you wanted, we could combo these lines into:

`return url.match(URL_PATTERN_MATCH);`

adamjarling: Also, can we add a quick unit test for this function in `assets/js/services/helpers.test.js`

adamjarling (changes requested):
Looks great.  A few small requests...   And if it's possible to add a RelatedURL unit test, that'd be great.  Sometimes writing tests for these field array components is complicated, but if this is not super complicated, let's get a test on it so we can know it's solid moving forward.

assets/js/services/helpers.js
katdivyareddy10: `url.match()` returns null when empty. I updated the code to a ternary instead of using a `const`. Added a test case under helpers.test.js for this function and possible outcomes.

katdivyareddy10:
RelatedURL uses react-hook-form and to test the error-messages, we need someway to submit a form from `WorkTabsAbout` component. As per our conversation, this functionality must be tested from `WorkTabsAbout` component. Also there's an issue [924](https://github.com/nulib/repodev_planning_and_docs/issues/924 ) to research best practices for testing such components. 

I have however added tests for `RelatedURL` component with a HigherOrder component.

assets/js/services/helpers.test.js
```diff
@@ -11,6 +12,17 @@ it("should escape double quotes", () => {
   expect(expected).toMatch(actual);
 });
 
+it("should validate URL pattern", () => {
+  expect(isUrlValid("htttp://northwestern.edu")).toBe(false);
```
adamjarling, line 16: This is a valid url (without the `www`).  Can we adjust the regex to allow domains of this type?

adamjarling (changes requested):
Looks good.  Small condition check in the regex... think you can find a regex pattern which allows the condition mentioned below?

assets/js/services/helpers.test.js
katdivyareddy10: This is invalid as it reads `htttps` and not `https`, there's an extra `t`, otherwise, it should be valid

kdid (approved):

assets/js/services/helpers.test.js
adamjarling: Ooops, I'm blind.  Sorry:)

PR #1198, opened 2020-09-17
Add required logic to Descriptive and Administrative metadata fields.…

adamjarling:
…  Also added the logic where a Work cannot be published unless it's part of a Collection.

Required fields based on this Sheet:
https://docs.google.com/spreadsheets/d/1PBvp9FJAvG6JIP6RZwmiZSDhgYLQ6VRSknUnkodEH8Q/edit#gid=1424571002

image


assets/js/screens/Work/Work.jsx
```diff
@@ -158,6 +160,16 @@ const ScreensWork = () => {
...
                     </dd>
                   </dl>
+
+                  {!data.work.collection && (
+                    <p className="notification has-text-centered">
+                      <FontAwesomeIcon icon="exclamation-triangle" /> A work
+                      must belong to a Collection in order to be published.
+                      Click on the{" "}
+                      <a href="#tab-administrative">Administrative tab</a>{" "}
+                      below, to add this work to a Collection.
```
katdivyareddy10, line 170: `#tab-administrative` navigates to About tab and does not open Administrative Tab.

katdivyareddy10 (commented):
Looks good except for the link to the tab.

assets/js/screens/Work/Work.jsx
adamjarling: Yea, it's not supposed to open the Administrative tab, but rather just scroll down to that area of the page.  Maybe it's confusing to add the link here?  I was trying to get fancy:)

katdivyareddy10: It's only confusing as on my browser, it is scrolling up to `Core Metadata` box and not scrolling to `Tabs`

adamjarling: Ah, ok I think I know a fix to make it a little cleaner....

adamjarling: The navbar header is getting in the way, and there's nothing else close above in the UI to attach an `id` to which we can use as an anchor target, so just removing for now.  Less confusion is probably better.

katdivyareddy10 (approved):

PR #1221, opened 2020-09-22
Add error message for batchEdit confirmation with no edits

katdivyareddy10:
Screen Shot 2020-09-22 at 2 27 57 PM

### DIFF
--- a/assets/js/components/BatchEdit/About/Confirmation.jsx
+++ b/assets/js/components/BatchEdit/About/Confirmation.jsx
@@ -6,7 +6,10 @@ import { toastWrapper } from "../../../services/helpers";
 import { BATCH_UPDATE } from "../batch-edit.gql";
 import { useMutation } from "@apollo/client";
 import BatchEditConfirmationTable from "./ConfirmationTable";
-import { removeLabelsFromBatchEditPostData } from "../../../services/metadata";
+import {
+  CONTROLLED_METADATA,
+  removeLabelsFromBatchEditPostData,
+} from "../../../services/metadata";
 import { useHistory } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


assets/js/components/BatchEdit/About/Confirmation.jsx
```diff
@@ -140,28 +140,37 @@ const BatchEditConfirmation = ({
...
-                  label="Confirmation Text"
-                  required
-                  data-testid="input-confirmation-text"
-                />
+          {!hasDeletes && !hasAdds ? (
+            <div className="notification is-white">
+              <p className="has-text-danger has-text-centered mb-3">
+                <FontAwesomeIcon icon="exclamation-triangle" />
+                <strong> NOTE: </strong> Please edit at least one variable to
+                submit Batch Edit Changes.
```
adamjarling, line 148: @katdivyareddy10 Can we update this text to read "Batch edit currently only supports the following form items (Contributor, next time, etc...)".   And can we make the text not the red "danger" color and wrap in a `<p className="notification">...</p>`?

adamjarling (changes requested):

assets/js/components/BatchEdit/About/Confirmation.jsx
katdivyareddy10: I made changes, but since `<ul>` cannot be descendant of `<p>`, I moved it to outside of p and wrapped everything in a  `<div className="notification">`

adamjarling: Looks great.

adamjarling (approved):

PR #1230, opened 2020-09-24
Replace the iterative work creation with a more fault tolerant GenServer

mbklein:
### Testing
1. Reset your dev stack and run meadow's DB/LDAP setup tasks
2. Run `mix meadow.seed_data` with a reasonable sized (50-100 works) batch
3. Kill it with `^C^C` partway through.
4. Start up the `phx.server`
5. Quickly browse to the ingest sheet to see its progress bar
6. Watch it climb to 100% and display the right number of works

If you want, you can kill it multiple times and let it start up again. There might be a 60 second pause while works that had been set to `processing` time out and go back to `pending`, but it should eventually, reliably get to 100%.


kdid:
We chatted about this. Requesting documentation PR to review at same time.

lib/meadow/ingest/work_creator.ex
```diff
@@ -0,0 +1,209 @@
...
+        ingest_sheet: ingest_sheet.id,
+        ingest_sheet_row: row_num
+      })
+    end)
+
+    work
+  end
+
+  @doc """
+  Ingest a single work and update its status inside a single atomc transaction
```
bmquinn, line 124:
```suggestion
  Ingest a single work and update its status inside a single atomic transaction
```

bmquinn (approved):

PR #1262, opened 2020-10-01
Add multi edit items to BatchProvider, and now wired up to edit multi…

adamjarling:
…ple single works via the Search screen

image

image


katdivyareddy10 (commented):
This looks great! Can we also place this header at the bottom of page? It will make it easy for navigation
Screen Shot 2020-10-02 at 8 08 39 AM

adamjarling:
@katdivyareddy10 Why don't we get this up on staging, and see what users would prefer?  I originally had it at the bottom, then just felt (to me) like it was getting lost when moving from Work to Work.  I'm fine with it either way...probably a good candidate for feedback.

katdivyareddy10 (approved):

PR #1263, opened 2020-10-01
Index model name as “name” instead of “title”

mbklein:
Fixes a longstanding bug in collection indexing and adds tests for all other indexable types


katdivyareddy10 (approved):
This is great, thank you for fixing this so quickly!

PR #1276, opened 2020-10-05
767 mock api fileset ordering

adamjarling:
Note this can be a partial PR, which uses mocked file sets in the current state so we can see the behavior of dragging file sets around.  I can't import DC seed data into Meadow for some reason, and don't have any Works w/ multiple file sets from the WWII Collection.

There's a potential shape for the payload object in the ticket.  It depends whether we want to handle updates individually, or send file set updates in a batch manner as a big array of objects.

Another update in this PR is that we can now reference local `import`s on the client side w/o the relative path pattern which is getting increasingly a pain.   So there is an alias set up `@js`, which points to the `assets/js` relative path.

So if you wanted to import a component, this now works from anywhere in the client-side of the app and will make our lives easier if we're shifting components around:

`import WorkTabsAbout from "@js/components/Work/Tabs/About.jsx"`

If using VSCode, it should be wired up where the IDE will be aware of the new relative path for intellisense completion of directories and files.

image


katdivyareddy10 (approved):
Looks good, the drag and drop works smoothly!

PR #1293, opened 2020-10-07
Refactor and simplify uncontrolled field batch updates

mbklein:
Merge all uncontrolled fields at once instead of iterating over each one
Separate out fields that should be replaced from those being appended

### DIFF
--- a/assets/yarn.lock
+++ b/assets/yarn.lock
@@ -27,7 +27,31 @@
     core-js "2.6.0"
     zen-observable "0.8.11"
 
+"@apollo/client@^3.2.1":
+  version "3.2.1"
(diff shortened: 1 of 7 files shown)


mbklein:
So the front end will now pass four parameters:

`query`: The Elasticsearch query to match
`delete`: The map of controlled fields and values to *delete* from all matching records
`add`: The map of controlled _and_ uncontrolled fields and values to *append* to all matching records
`replace`: The map of uncontrolled fields and values to *replace* on all matching records

To nil out or empty an uncontrolled field, set its value to `null` or `[]` in the `replace` parameter.

bmquinn (approved):
👍

PR #1344, opened 2020-10-15
ProjectList back button bug fix

divyareddynu:
In `projectList.map` projectList is sometimes empty if the query `GET_PROJECT` is pending. Verifying `projectList` exists before looping through it takes away the bug.

### DIFF
--- a/assets/js/components/Project/List.jsx
+++ b/assets/js/components/Project/List.jsx
@@ -107,39 +107,40 @@ const ProjectList = () => {
           </tr>
         </thead>
         <tbody>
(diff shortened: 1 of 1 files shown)


adamjarling:
@divyareddynu Can you add some details in the description above of what the fix was?   It's not immediately clear to me from looking at the code.

divyareddynu:
> @divyareddynu Can you add some details in the description above of what the fix was? It's not immediately clear to me from looking at the code.

Just added it.

adamjarling (approved):

PR #1347, opened 2020-10-15
View Projects button for Project page

divyareddynu:
Screen Shot 2020-10-15 at 10 05 05 AM


adamjarling:
@divyareddynu So there should only be one "primary" button in a button group.  Can you make the less priority button a "regular" button instead of primary?

divyareddynu:
> @divyareddynu So there should only be one "primary" button in a button group. Can you make the less priority button a "regular" button instead of primary?

How does this look? 
Screen Shot 2020-10-15 at 10 29 15 AM

adamjarling (approved):

PR #1349, opened 2020-10-15
1124 batch edit multi value items

adamjarling:
This PR introduces a new component `assets/js/components/UI/Form/BatchFieldArray.jsx` which adds extra functionality to the `FieldArray` component.  
- User can select to make the field array form items "Replace" items
- Or user can do a replace all, which appears in the UI as "Remove...."

This follows the convention of the new API structure for `batchUpdate` mutation.

It also refactors the Batch Edit confirmation screen modal component to:
- Simplify the UI
- Handle non-controlled-term form item values

This turned out to be a lot of work, which is blocking further UI work, so I'd like to get this merged, w/o tests.  Creating another issue now to apply unit tests to all code introduced in this PR.

Lastly, I created a new Metadata Fields mapping object, which allows us to create different arrays of metadata form fields in different which allow us to loop through rendering, and reduce code duplication.  It's not complete, but a start.

`assets/js/services/metadata.js`

image

image

image


kdid (changes requested):
Remove `legacy_identifier` and `identifier`. They are not supported in the API (they are unique fields so cannot be updated via batch.)

adamjarling:
@kdid Removed Identifier and Legacy Identifier in latest commit.

kdid (approved):

PR #1358, opened 2020-10-16
BatchEdit for collection

divyareddynu:
For collection, it's currently displaying ID, i would however like to display `Collection title` by fetching it using `GET_COLLECTION` query in `BatchConfirm` page. So the setup of adding collection as an object to batchreplace won't work as it accepts only string per graphQL. Let me know your thoughts and I will modify code accordingly.

Screen Shot 2020-10-16 at 8 07 17 AM

---

Screen Shot 2020-10-16 at 6 08 42 AM
Screen Shot 2020-10-16 at 6 08 50 AM
Screen Shot 2020-10-16 at 6 31 23 AM

### DIFF
--- a/assets/js/components/BatchEdit/About/About.jsx
+++ b/assets/js/components/BatchEdit/About/About.jsx
@@ -30,6 +30,7 @@ const BatchEditAbout = () => {
   const [batchReplaces, setBatchReplaces] = useState({
     descriptiveMetadata: {},
   });
+  const [batchCollection, setBatchCollection] = useState({});
 
   const batchDispatch = useBatchDispatch();
 
@@ -100,8 +101,12 @@ const BatchEditAbout = () => {
     setBatchDeletes(deleteReadyItems);
     setBatchReplaces({
       descriptiveMetadata: { ...replaceItems, ...multiValues.replace },
-      collectionId: currentFormValues["collectionId"],
     });
+
+    Object.keys(currentFormValues["collection"]).length > 0
+      ? setBatchCollection(JSON.parse(currentFormValues["collection"]))
+      : setBatchCollection({});
+
     setIsConfirmModalOpen(true);
   };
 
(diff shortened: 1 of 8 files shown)


adamjarling:
@divyareddynu Slack me when you get a chance.  I want to make sure I'm understanding this.

assets/js/components/BatchEdit/About/CoreMetadata.jsx
```diff
@@ -45,6 +52,25 @@ const BatchEditAboutCoreMetadata = ({ ...restProps }) => {
           label="Alternate Title"
         />
       </div>
+      <div className="column is-full">
+        <UIFormField label="Collection">
+          <UIFormSelect
+            isReactHookForm
+            name="collectionId"
```
adamjarling, line 59: Semantically a form `name` should just describe the data.  We don't need the "Id".  So let's make `name="collection"` instead.

adamjarling:
@divyareddynu Ok, so I think what you're saying is we need to pull through the title of the Collection to the final Confirmation component, and currently we're just saving the Collection `id` as the form select value?

If so, I'd suggest the following:

1. As the `<option value="...">` for `value` let's save an object which contains the Collection `id` and `title`.   You might need to `JSON.stringify` this value.
2. In the About component, let's create a new `useState()` variable for `batchCollection`, (which holds the object), and pass that value to `Confirmation` component.
3. In the `Confirmation` component, we just add `collectionId: [PULL OUT COLLECTION ID HERE]`

Make sense?

divyareddynu:
@adamjarling Here's an updated version 
Screen Shot 2020-10-16 at 1 46 58 PM

assets/js/components/BatchEdit/About/About.jsx
```diff
@@ -100,8 +101,12 @@ const BatchEditAbout = () => {
     setBatchDeletes(deleteReadyItems);
     setBatchReplaces({
       descriptiveMetadata: { ...replaceItems, ...multiValues.replace },
-      collectionId: currentFormValues["collectionId"],
     });
+
+    Object.keys(currentFormValues["collection"]).length > 0
```
adamjarling, line 106: No big deal, but another common way to make this check could be:  `currentFormValues.hasOwnKey("collection")`

divyareddynu: @adamjarling  I used only to eliminate the worst of cases where collection is empty string like:

Screen Shot 2020-10-19 at 10 52 00 AM

adamjarling (approved):

PR #1359, opened 2020-10-16
Update homepage

adamjarling:
Added 2 big actions, along with some mock charts to see if anything clicks on what we could display.

image


divyareddynu (approved):
This is super cool!

PR #1368, opened 2020-10-20
Send validation notifications periodically instead of as a flood

mbklein:
Create IntervalTask for periodic task calls
Refactor existing periodic tasks using IntervalTask
Implement ValidationNotifier using IntervalTask
Split WorkRedriver off from WorkCreator

### DIFF
--- a/coveralls.json
+++ b/coveralls.json
@@ -7,6 +7,7 @@
     "lib/meadow_web/channels/user_socket.ex",
     "lib/meadow_web/router.ex",
     "lib/meadow/application.ex",
+    "lib/meadow/data/works/batch_functions.ex",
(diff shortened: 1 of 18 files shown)


mbklein:
We can improve the dip in test coverage after we send the code to staging for testing.

kdid:
So, is this on the 15 second interval we talked about?

mbklein:
I had whiffed on the specific delay we discussed, but have just force pushed a change. Validation and Ingest progress updates are now at 15 second intervals. I think we can bring the number _way_ down (to like 500 or 1000 milliseconds each) without harm, but this is a good starting place to check the concept.

kdid (approved):

PR #1378, opened 2020-10-20
Increase Progress.initialize_entries/1 timeout to :infinity

mbklein:
Make ReleaseTasks.reset!/0 pause and resume IntervalTask processes before destroying the database

### DIFF
--- a/lib/meadow/ingest/progress.ex
+++ b/lib/meadow/ingest/progress.ex
@@ -87,11 +87,14 @@ defmodule Meadow.Ingest.Progress do
   def initialize_entries(entries) do
     timestamp = DateTime.utc_now()
 
-    Repo.transaction(fn ->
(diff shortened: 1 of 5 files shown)


bmquinn (commented):
If I run `Meadow.ReleaseTasks.reset!()` twice, I see warnings for each migration like:
```
warning: redefining module Meadow.Repo.Migrations.LoadUuidModule (current version defined in memory)
  _build/dev/lib/meadow/priv/repo/migrations/0000000000000_load_uuid_module.exs:1
```
Think it's a problem that needs to be addressed or something we can ignore?

mbklein:
I think we can ignore that. Running `reset!()` twice in the same session seems weird to me anyway.

kdid (approved):

PR #1386, opened 2020-10-22
1159 admin tab batch edit

divyareddynu:
Closes [1137](https://github.com/nulib/repodev_planning_and_docs/issues/1137)

Screen Shot 2020-10-22 at 9 48 39 AM


assets/js/services/metadata.js
```diff
@@ -362,34 +395,74 @@ export function prepRelatedUrl(items = []) {
...
-        key
-      ].map((item) => {
-        // Regular string value
-        if (typeof item !== "object") {
-          return item;
+    batchAdds.descriptiveMetadata &&
+      Object.keys(batchAdds.descriptiveMetadata).forEach((key) => {
+        returnObj.add.descriptiveMetadata[key] = batchAdds.descriptiveMetadata[
+          key
+        ].map((item) => {
```
adamjarling, line 423: @divyareddynu Looks like we're re-using this code block (lines 423 - 432) three separate times.  Just to make our life easier in the future to maintain this code, how about we pull this out into it's own helper function?

adamjarling (changes requested):
Getting really close and looking good.  There's so much code, just trying to figure out ways to thin it out where possible:)  I think make this update to save us ~ 50 lines, then looking ready to merge.

kdid (approved):

PR #1414, opened 2020-10-28
Ingest sheet validation use status field to show/hide preview items o…

adamjarling:
…r errors

Previously this was relying upon a percentage value from progress bar data to show results, but now using `status` instead.

All in all, cleaned up the code and thinned out the data flowing through the components in Validation, up until the point where a user hits "Approve".    Moving onto that next.


divyareddynu:
Looks good but I'm having issues getting past validation. I tried with 3 different sheets.
Screen Shot 2020-10-29 at 8 50 38 AM
Screen Shot 2020-10-29 at 8 50 47 AM

kdid (approved):

adamjarling:
@divyareddynu Yea, investigating that part next to see what's going on.

PR #1418, opened 2020-10-29
Bug fix for crossbrowser select options

divyareddynu:
The font-family is not supported in some browsers and it is a known issue, per https://bugzilla.mozilla.org/show_bug.cgi?id=1536148 which is still OPEN. 
I've successfully imposed it and was able to test on multiple browsers using browserstack with a custom font that bulma uses for its select options. I couldn’t get IE/Edge fired up to test it.

Screen Shot 2020-10-29 at 8 39 19 AM


adamjarling (changes requested):
@divyareddynu Yea, looks like that is Bulma's default font.  Why don't we do 2 quick things:

1. Let's trim down the `font-family` definition to use the primary font our application is using first, followed by `BlinkMacSystemFont, -apple-system, Arial, sans-serif`.  We don't need all the rest
2. Since we're reusing this, I'd suggest let's put into a `class` defined in `assets/styles/scss/_base.scss`

adamjarling (approved):

PR #1422, opened 2020-10-29
Bug fix for osd viewer reload

divyareddynu:
Just by introducing`key` on the viewer component, react re-renders it as key changes for every work.


adamjarling (approved):
Best fix ever!

PR #1428, opened 2020-10-30
Pin react-hook-form to v6.9.2, as there appears to be a bug in their …

adamjarling:
…reset() method.  Also upgraded React to v17.

I couldn't find a solution to this after burning 2 hours on it, so just reverting back to v6.9.2   It could either be a bug in React Hook Form's `reset()` method, or a collision with how we're implementing FieldArrays, then calling `reset()`.   We can look into it later after the next release in a week or so and see if it resolves itself.

To test, make sure to do a `yarn install` before testing in the UI.


divyareddynu (approved):
Works as expected, batchEdit now submits and returns to /search page.

PR #1437, opened 2020-11-02
Add date_created to API and Ecto schema

kdid:
- Adds `date_created` to work descriptive metadata
- Creates custom Ecto type for `date_created`
  - DB format: `%{edtf_date: "1975-07-01", humanized_date: "Tue, 01 Jul 1975"}`
  - for now there is no actual validation (since that is represented in other tickets). (any string will "humanize" as "Tue, 01 Jul 1975")
- Adds `date_created` to GraphQL API (work queries and mutations and batchUpdate mutation)

Screen Shot 2020-11-02 at 1 04 54 PM

*** Requires a `down -v` locally or a database reset on staging....

### DIFF
--- a/lib/meadow_web/schema/types/data/batch_types.ex
+++ b/lib/meadow_web/schema/types/data/batch_types.ex
@@ -82,6 +82,7 @@ defmodule MeadowWeb.Schema.Data.BatchTypes do
     field :box_number, list_of(:string)
     field :caption, list_of(:string)
     field :catalog_key, list_of(:string)
+    field :date_created, list_of(:edtf_date_input)
     field :description, list_of(:string)
     field :folder_name, list_of(:string)
     field :folder_number, list_of(:string)
--- a/lib/meadow/data/schemas/types/edtf_date.ex
+++ b/lib/meadow/data/schemas/types/edtf_date.ex
@@ -1,0 +1,47 @@
+defmodule Meadow.Data.Types.EDTFDate do
+  @moduledoc """
+  Ecto.Type for converting between edtf string and humanized date
+  """
+
+  use Ecto.Type
+  alias Meadow.Data.Works
+
+  def embed_as(:json), do: :dump
+
+  def type, do: :map
+
+  def cast(edtf_date), do: humanize(edtf_date)
+
+  def load(edtf_date), do: humanize(edtf_date)
+
+  def dump(nil), do: nil
+
+  def dump(%{edtf_date: edtf_date, humanized_date: humanized_date}),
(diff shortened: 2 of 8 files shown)


mbklein (approved):

adamjarling:
Looks good here.  @kdid Assume those string formats in the API screenshot for dates can be created via the EDTF JS package we've been talking about?

kdid:
Yeah, between the edtf.js package and possibly the additional humanize package. But the string you need to submit is not generated. The raw EDTF date string is something that will be entered by users. If it validates you can get a humanized form of it.... But you don't need to actually submit the humanized form. It is looked up on the backend.

lib/meadow_web/schema/types/data/batch_types.ex
```diff
@@ -82,6 +82,7 @@ defmodule MeadowWeb.Schema.Data.BatchTypes do
     field :box_number, list_of(:string)
     field :caption, list_of(:string)
     field :catalog_key, list_of(:string)
+    field :date_created, list_of(:edtf_date_input)
```
bmquinn, line 85: Happy to hear your thoughts especially if I'm missing something, but my take is that `:edtf_date_input` is probably overly complicated if all we're going to input is a string here. Would probably just go with `list_of(:string)` and keep the `list_of(:edtf_date_entry)` on the `:work_descriptive_metadata` object.

kdid: I was thinking it might make it easier on the front end to send/display an object (just w/out one of the fields on the input case) ?

bmquinn: It was more the input object itself that threw me a little bit so I had to dig into the docs, unless we need something more complicated in the frontend `createWork(descriptiveMetadata:{dateCreated:"?2016-?05-31"},...)` seems more natural to me than `createWork(descriptiveMetadata:{dateCreated:{edtfDate:"?2016-?05-31"}},...)`. I could get used to the latter, but just throwing out my $.02.

kdid: I do hear you. And I agree with what you are saying - it's just I thought it might be easier for the front end (similar to what we do for coded terms) that they get back from the API  queries and mutations: `dateCreated: {edtf_date: "1975-07-01", humanized_date: "Tue, 01 Jul 1975"}` and then they can turn around and basically send the input object in the same format (minus the humanized) w/out having to reformat it.  -- Also not totally sure how it will all play out - but it's possible we might be able to just optionally take the humanized form from the front end/API and save a lookup on the backend if it's provided.....

I'd like to leave it for now and see how the front end evolves. If it turns out that it's overkill, switch it to a string.....

PR #1439, opened 2020-11-03
bug fix for right statement batchedit

divyareddynu:
Screen Shot 2020-11-02 at 6 08 35 PM

### DIFF
--- a/assets/js/components/BatchEdit/About/CoreMetadata.jsx
+++ b/assets/js/components/BatchEdit/About/CoreMetadata.jsx
@@ -5,7 +5,6 @@ import UITagNotYetSupported from "../../UI/TagNotYetSupported";
 import UIInput from "../../UI/Form/Input";
 import UIFormField from "../../UI/Form/Field";
 import UIFormBatchFieldArray from "../../UI/Form/BatchFieldArray";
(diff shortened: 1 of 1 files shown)


assets/js/components/BatchEdit/About/CoreMetadata.jsx
```diff
@@ -82,18 +81,24 @@ const BatchEditAboutCoreMetadata = ({ ...restProps }) => {
...
+                      scheme: "RIGHTS_STATEMENT",
+                      label: item.label,
+                    })}
+                  >
+                    {item.label}
+                  </option>
+                ))}
+            </select>
+          </div>
         </UIFormField>
```
adamjarling, line 102: Can you explain how this fixes the bug?   Would the same effect be achieved by prepping an `options` array and feeding it into `UIFormSelect`, or is there something with `UIFormSelect` that gets in the way?

divyareddynu: For `UIFormSelect`, there is not an easy way to mention custom values for options, like in the case of codelists, we need to provide ID: RIGHTS_STATEMENT and also clear out label field that we use in batchConfirmation screen.

adamjarling (approved):

PR #1453, opened 2020-11-04
Clean up Work and Batch Edit form UI

adamjarling:
This tightens up all the white space in Work and Batch Edit forms.  Especially on the Controlled item boxes.   I also made the buttons regular size and the `isLight` color so they stick out a bit more and the user can see what the action is?

image

image

image

### DIFF
--- a/assets/js/components/BatchEdit/About/ControlledMetadata.jsx
+++ b/assets/js/components/BatchEdit/About/ControlledMetadata.jsx
@@ -39,10 +39,10 @@ const BatchEditAboutControlledMetadata = ({ ...restProps }) => {
 
   return (
     <div data-testid="controlled-metadata" {...restProps}>
-      <ul>
+      <ul className="columns is-multiline">
         {!codeLists.isLoading &&
           CONTROLLED_METADATA.map(({ label, name, scheme }) => (
-            <li key={name} className="mb-5" data-testid={name}>
+            <li key={name} className="column is-half mb-5" data-testid={name}>
               <UIFormField label={label}>
                 <UIFormControlledTermArray
                   authorities={codeLists.authorityData.codeList}
--- a/assets/js/components/BatchEdit/ModalRemove.js
+++ b/assets/js/components/BatchEdit/ModalRemove.js
@@ -68,7 +68,9 @@ export default function BatchEditAboutModalRemove({
         <div className="box">
           <header>
             <h3 className="title">{currentRemoveField.label}</h3>
-            <h4 className="subtitle">Batch remove the following entries</h4>
+            <h4 className="subtitle">
+              Select values to remove from all batch Works
+            </h4>
           </header>
           <div className="my-4">
             {candidateList.map((item) => (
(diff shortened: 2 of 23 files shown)


assets/js/components/BatchEdit/ModalRemove.js
```diff
@@ -90,8 +92,8 @@ export default function BatchEditAboutModalRemove({
           </div>
           <footer>
             <div className="buttons is-right">
-              <Button isText onClick={closeModal}>
-                Close
+              <Button isLight onClick={closeModal}>
+                Save &amp; close
```
divyareddynu, line 96: Can we make this capital `Save & Close`?

assets/js/components/BatchEdit/About/ControlledMetadata.jsx
```diff
@@ -39,10 +39,10 @@ const BatchEditAboutControlledMetadata = ({ ...restProps }) => {
 
   return (
     <div data-testid="controlled-metadata" {...restProps}>
-      <ul>
+      <ul className="columns is-multiline">
         {!codeLists.isLoading &&
           CONTROLLED_METADATA.map(({ label, name, scheme }) => (
-            <li key={name} className="mb-5" data-testid={name}>
+            <li key={name} className="column is-half mb-5" data-testid={name}>
```
divyareddynu, line 45: Can we make it two lines with less margin/padding as below always instead of space-permitting?

Screen Shot 2020-11-04 at 11 21 48 AM

divyareddynu (commented):
Really like the changes and the work screen card with more data. Minor thoughts here and there.

kdid:
Like the bigger buttons for add and remove controlled terms. 

My two cents would be - 

"View and remove contributors" instead of "Remove entries in contributor"..... etc

And "Uneditable Metadata" is a weird word, I think I'd just leave that header off and have those id's just sitting by themselves in that box at top.....

assets/js/components/BatchEdit/ModalRemove.js
adamjarling: @divyareddynu I guess I was looking at Github here as an example for the buttons and how they handle capitilization.  I kind of like the style of having buttons read as mini sentences, if that's ok by you.

assets/js/components/BatchEdit/About/ControlledMetadata.jsx
adamjarling: Yea, good call.   Those are buttons (text buttons, so not totally sure how it'd work but I'll see if there's something we can do.

assets/js/components/BatchEdit/ModalRemove.js
divyareddynu: Cool, marking this as resolved.

assets/js/components/BatchEdit/About/ControlledMetadata.jsx
divyareddynu: Thank you!

adamjarling:
Here's an update to the Controlled Term list display now @divyareddynu 

image

Made them a flexbox column.  It will probably stack better this way with 2 columns across.

divyareddynu:
This is great! Thank you for making the change!

kdid (approved):

PR #1462, opened 2020-11-05
Replace real input values with placeholders for non-controlled term f…

adamjarling:
…ield array form metadata items

Now looks like this for non-controlled term form elements:

image

### DIFF
--- a/assets/js/components/Work/Tabs/About.jsx
+++ b/assets/js/components/Work/Tabs/About.jsx
@@ -28,9 +28,51 @@ import {
 } from "../../../services/metadata";
 import UIError from "../../UI/Error";
 
+function prepFormData(work) {
+  const { descriptiveMetadata } = work;
+  let resetValues = {};
+  let controlledTermResetValues = {};
+
+  // Convert data data from the API for a shape the form (and React Hook Form) want.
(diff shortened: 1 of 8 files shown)


assets/js/components/Work/Tabs/About.jsx
```diff
@@ -314,4 +321,4 @@ WorkTabsAbout.propTypes = {
   work: PropTypes.object,
 };
 
-export default WorkTabsAbout;
+export default React.memo(WorkTabsAbout);
```
divyareddynu, line 324: Cool!

assets/jsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "compilerOptions": {
-    "baseUrl": "./",
+    "jsx": "react", // Allow command-click open components in a new file in VSCode
```
divyareddynu, line 3: Thank you for this! I tried with your branch on a UIFormField and it didn't work as expected. Is this only for some files?

divyareddynu (approved):

assets/jsconfig.json
adamjarling: @divyareddynu Try restarting VSCode?   I'll double check too.

PR #1475, opened 2020-11-09
Support batch Published updates for publish and unpublish all items i…

adamjarling:
…n the UI

A user can now publish or unpublish Works in batch edit mode.  The checkboxes are mutually exclusive.

image


kdid:
Batch published works great. I did notice then when I tried to "Unpublished" a single work it wouldn't let me because it didn't belong to a collection. Unpublishing shouldn't require collection membership....



Screen Shot 2020-11-09 at 2 45 37 PM

adamjarling:
@kdid Good catch.  Just pushed up a commit to make sure the "Unpublish" button is never disabled.

PR #1485, opened 2020-11-11
Add BatchEdit and Edit components for EDTF type of input

divyareddynu:
Screen Shot 2020-11-11 at 8 30 39 AM
Screen Shot 2020-11-11 at 8 31 01 AM


adamjarling (changes requested):
Looking good @divyareddynu   A few things...

I think we need better messaging for a user on what a valid date is, or an indication this is an EDTF date with a link maybe to some EDTF documentation?   For example, I get this and have to idea what to do as a user:

image

Let's use Bulma's `help` class to add some more info for the user.  Something like:

"This is and EDTF date value.  View spec"   And "View spec" links to: https://www.loc.gov/standards/datetime/

Also, since we'll most likely not re-use the `UIFormBatchEDTFDate`, I'm wondering if it'd make more sense for the component to live in the `BatchEdit` directory?   Thoughts?

adamjarling:
Can we also pull up the "Rights Statement" form element to be in same row as "Date Created" to conserve space?

image

assets/js/components/BatchEdit/About/CoreMetadata.jsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
-import UITagNotYetSupported from "../../UI/TagNotYetSupported";
+import UIFormBatchEDTFDate from "../../UI/Form/BatchEDTFDate";
```
adamjarling, line 3: Anything new we add, let's make sure we're always using the relative import pattern to save us time down the road:   `@js/components/UI/Form/BatchEDTFDate`

assets/js/components/UI/Form/BatchEDTFDate.jsx
```diff
@@ -0,0 +1,135 @@
...
+import { isEDTFValid } from "../../../services/helpers";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { Button } from "@nulib/admin-react-components";
+
+const UIFormBatchEDTFDate = ({
+  name,
+  label,
+  type = "text",
+  required,
+  ...passedInProps
```
adamjarling, line 15: How about from now on, since this seems to be a more frequent pattern, we always title JS "rest" props, `restProps`, instead of `passedInProps`?

assets/js/services/helpers.js
```diff
@@ -42,6 +43,15 @@ export function getClassFromIngestSheetStatus(status) {
   return "";
 }
 
+export function isEDTFValid(edtfString) {
+  try {
+    edtf(edtfString).edtf;
+    return true;
+  } catch (e) {
+    return false;
```
adamjarling, line 51: So the `edtf` method on line 48 throws an Error if its invalid?   Does it give any type of indication of what went wrong?

assets/js/services/metadata.js
```diff
@@ -241,13 +242,13 @@ export function getBatchMultiValueDataFromForm(currentFormValues) {
   );
 
   for (const key of formMultiOnly) {
+    let rootName = key.split("--")[0];
```
adamjarling, line 245: Nice, this is easier to understand.

assets/js/components/UI/Form/EDTFDate.jsx
```diff
@@ -0,0 +1,100 @@
...
+  const { control, errors, register } = useFormContext();
+  const { fields, append, remove } = useFieldArray({
+    control,
+    name, // Metadata item form name
+    keyName: "useFieldArrayId",
+  });
+
+  return (
+    <div data-testid="dateCreated-wrapper">
+      <fieldset>
```
adamjarling, line 19: Since there's so much code duplication between here and `UIFormFieldArray` component, is there a way we can use that component?    

Let's think about why we're not using it, and what would need to be adjusted in the `UIFormFieldArray` component to allow its usage?

adamjarling: Is it just the custom validation?  If so, maybe we build that into`UIFormFieldArray`

adamjarling:
We should continue the pattern of providing placeholder for any new items added in Field Array items:

image

adamjarling:
We should make the property `metadataItem` instead of `edtf`.  `metadataItem` is the default holder for single value Non-Controlled field array items.   I think this will get resolved if we use the `UIFormFieldArray` component too though...

assets/js/components/Work/Tabs/About.jsx
```diff
@@ -108,12 +109,14 @@ const WorkTabsAbout = ({ work }) => {
...
     // updated.
     let currentFormValues = methods.getValues();
+
     const { title = "" } = currentFormValues;
     let workUpdateInput = {
       descriptiveMetadata: {
         alternateTitle: prepFieldArrayItemsForPost(
           currentFormValues.alternateTitle
         ),
+        dateCreated: currentFormValues.dateCreated || [],
```
adamjarling, line 119: Here we want to pass to `prepFieldArrayItemsForPost` function to convert from an array of objects to array of strings.

assets/js/services/metadata.js
```diff
@@ -401,7 +402,19 @@ export function prepRelatedUrl(items = []) {
  * @param {Object} item
  * @returns {Object}
  */
-export function deleteKeyFromObject(item) {
+export function deleteKeyFromObject(item, key) {
+  if (key === "dateCreated") {
```
adamjarling, line 406: Do we need this block of code?   It seems the function should be generic and accounting for a single value `dateCreated` within this function might not be the best spot for it, if the code block is really even needed.  I can't tell why we're calling it...

adamjarling: Ah, now I see why that code is there.  Because of the API:

image

I get it, but think we should move this unique scenario, which converts an array of string values, into an "edtf object" creator function.  

Maybe we code for that check in the `handleBatchEditConfirm()` function in the `Confirmation` component, and call it after we have a value for `cleanedPostValues`.

assets/js/components/UI/Form/EDTFDate.jsx
divyareddynu: I agree that we should keep DRY code.

assets/js/services/helpers.js
divyareddynu: The what went wrong is not indicated as the package throws garbage text if the date is invalid.

assets/js/components/Work/Tabs/About/CoreMetadata.jsx
```diff
@@ -1,20 +1,23 @@
...
 import PropTypes from "prop-types";
-import UITagNotYetSupported from "../../../UI/TagNotYetSupported";
 import UIInput from "../../../UI/Form/Input";
 import UIFormField from "../../../UI/Form/Field";
 import UIFormSelect from "../../../UI/Form/Select";
 import UIFormFieldArray from "../../../UI/Form/FieldArray";
 import UIFormFieldArrayDisplay from "../../../UI/Form/FieldArrayDisplay";
 import UICodedTermItem from "../../../UI/CodedTerm/Item";
 import { useCodeLists } from "@js/context/code-list-context";
+import { isEDTFValid } from "../../../../services/helpers";
```
adamjarling, line 10: As long as we're making some updates to this commit, let's use the relative import syntax as seen in line 9.  All those `../../` are making my brain hurt:)

assets/js/components/Work/Tabs/About/CoreMetadata.jsx
```diff
@@ -1,20 +1,23 @@
...
 
 const WorkTabsAboutCoreMetadata = ({
   descriptiveMetadata,
   isEditing,
   published,
 }) => {
   const codeLists = useCodeLists();
+  const EDTFValidateFn = (value) => {
+    return isEDTFValid(value) || <span>Please enter raw EDTF date. </span>;
```
adamjarling, line 19: Looks like we can add a `<a href="" />` element here.  Can we add the link to the EDTF Library of Congress site?

assets/js/services/metadata.js
```diff
@@ -410,6 +411,14 @@ export function deleteKeyFromObject(item) {
   return itemObj;
 }
 
+export function prepEDTFforPost(items) {
```
adamjarling, line 414: No biggie, but you can set a default param for `items` (`prepEDTFforPost(items = [])`) and then don't need the ternary in the code itself.

adamjarling (changes requested):
Looking great.  Some minor adjustments, but then I think it's ready for merge.

adamjarling (approved):

PR #1512, opened 2020-11-20
Use EFS storage for pyramid TIFF processor

mbklein:
* Add EFS filesystem and mount points
* Mount EFS filesystem on container as `/working`
* Allow Meadow to (optionally) use alternate working directory for TIFF derivatives
* Clean up config code

**NOTE**: The Terraform code in this PR needs to be applied before the Meadow changes are deployed

### DIFF
--- a/config/releases.exs
+++ b/config/releases.exs
@@ -70,18 +70,16 @@ config :meadow,
     target_url: get_required_var.("EZID_TARGET_BASE_URL")
   },
   environment: :prod,
-  pipeline_delay: System.get_env("PIPELINE_DELAY", "120000")
-
-config :meadow, ingest_bucket: get_required_var.("INGEST_BUCKET")
-config :meadow, preservation_bucket: get_required_var.("PRESERVATION_BUCKET")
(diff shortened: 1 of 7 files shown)


bmquinn:
This looks great, I had one question that wasn't clear to me looking over the code. I see where `Config.tiff_port_environment()` is passed to the port in `Meadow.Pipeline.Actions.CreatePyramidTiff`, but how does `pyramid.js` and/or `tempy.file()` know to use the configured working directory in the JavaScript?

mbklein:
Good question! I tried to get `tempy` to generate a filename in a specific directory, but it will only use the system temp directory pointed to by the POSIX-standard `TMPDIR` environment variable. The easiest way to make it use the mounted working volume without screwing things up for our dev boxes was to override `TMPDIR` in the environment passed to the port.

bmquinn:
Right on, I just dug through the `tempy` repo and it looks like they piggyback off of the Node.js `os.tmpdir()` method which appropriately handles the `TMPDIR` env var (see https://github.com/sindresorhus/tempfile/issues/12). Thanks 👍

bmquinn (approved):

PR #1521, opened 2020-11-23
Backend for Batch Dashboard

kdid:
- Backend for batch update dashboard and one at a time batch processing.
   - Batch update mutation creates a batch in the database with a `status` of "queued"
- `Meadow.BatchDriver` `IntervalTask` that looks for `queued` batches to start
   - `active` flag in `batches` table has a unique constraint on the `true` value to prevent more than one batch from running at once. 
   - Also looks for batches that are `active` = `true` (and `status` = `in_progress` for longer than the `@timeout` (which I randomly set at 6 minutes. Not sure if that's realistic or not. Need to test on staging)
- Adds a many to many relationship between `Batch` and `Work`  and indexes `batch_ids` on works (so batches can be used as a facet in the UI)
- Adds the `current_user` from the Absinthe context to the batch entry.
- Batches fail/rollback - all or nothing - and an error message is stored in the `error` field in the database.

**_Edited some existing migrations so will require a data reset!_**

Notes:
- There is a tad of inconsistency on validation/error handling behavior among different fields - I'll open an issue for cleanup (https://github.com/nulib/repodev_planning_and_docs/issues/1296)

### DIFF
--- a/lib/meadow/batch_driver.ex
+++ b/lib/meadow/batch_driver.ex
@@ -1,0 +1,38 @@
+defmodule Meadow.BatchDriver do
+  @moduledoc """
+  IntervalTask to kickoff queued batch update/delete jobs
+  """
+  import Ecto.Query, warn: false
+  require Logger
+
+  alias Meadow.Batches
+  alias Meadow.IntervalTask
+
+  use IntervalTask, default_interval: 30_000, function: :drive_batch
+
+  require Logger
+
+  @timeout 360
+
+  @doc """
+  If no batches are currently running, find the next one and start it
+  """
+  def drive_batch(state) do
+    with {:ok, count} <- Batches.purge_stalled(@timeout) do
+      if count > 0 do
+        Logger.info("Purging #{count} stalled #{Inflex.inflect("batch", count)} from queue")
+      end
+    end
+
+    case Batches.next_batch() do
+      nil ->
+        :noop
+
+      batch ->
+        Logger.info("Found next batch to start #{inspect(batch)}")
+        Batches.process_batch(batch)
+    end
+
+    {:noreply, state}
+  end
+end
--- a/lib/meadow/data/schemas/batch.ex
+++ b/lib/meadow/data/schemas/batch.ex
@@ -1,0 +1,82 @@
+defmodule Meadow.Data.Schemas.Batch do
+  @moduledoc """
+  Batch edit (update/delete) schema
+  """
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias Meadow.Data.Schemas.Work
+  alias MeadowWeb.Schema.ChangesetErrors
+
+  @statuses ~w(queued in_progress complete error)
+  @types ~w(update delete)
+
+  @primary_key {:id, Ecto.UUID, autogenerate: false, read_after_writes: true}
+  schema "batches" do
+    field :nickname, :string
+    field :status, :string, default: "queued"
+    field :user, :string
+    field :started, :utc_datetime_usec
+    field :type, :string
+    field :works_updated, :integer
+    field :query, :string
+    field :add, :string
+    field :delete, :string
+    field :replace, :string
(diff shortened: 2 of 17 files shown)


mbklein:
I like the overall design, especially the unique boolean field on the batches table. Fits in nicely with the “let the database do the work when possible” approach. I think there's a good opportunity here to use the `DatabaseNotification` module to build a pass-through GraphQL subscription to notify the front-end of status changes.

lib/meadow/batch_driver.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  @doc """
+  If no batches are currently running, find the next one and start it
+  """
+  def drive_batch(state) do
+    # todo - check for stale batches exceeding some extreme upper bounds timeout
+
+    case Batches.next_batch() do
+      nil ->
+        Logger.info("No batches queued")
```
bmquinn, line 27: It's a fine line with logging always, but my take here is this adds extra noise to the log. Would maybe just do `nil -> :noop` instead

lib/meadow/data/schemas/batch.ex
```diff
@@ -0,0 +1,48 @@
...
+      :type,
+      :works_updated,
+      :query,
+      :add,
+      :delete,
+      :replace,
+      :error,
+      :active
+    ])
+    |> unique_constraint(:active)
```
bmquinn, line 46: You could add `|> validate_inclusion(:status, ["complete", "error", "in_progress"])` for an extra data integrity check here as well

bmquinn (commented):
I think this looks great! I like the `unique_index` approach you took. Added a couple small comments as well.

mbklein (commented):
What is the purpose of the new `batch_ids` field in the work schema? I'm not sure batches (which represent an ephemeral process) need to be as part of the permanent representation of a work. I also can't find where they're used – just created and updated. If there needs to be a link between the batches table and the works affected by each batch, I'd rather have a separate table that links `batch_id` to `work_id` (or at least which keys on one and stores the other as an array).

test/meadow/batches_test.exs
```diff
@@ -201,21 +302,61 @@ defmodule Meadow.BatchesTest do
...
     end
+
+    test "purge_stalled/1 errors timed out batches stalled in_progress" do
+      replace =
+        Jason.encode!(%{
+          administrative_metadata: %{},
+          descriptive_metadata: %{title: ";sdakjf;alsdkjf;aksjdf"}
+        })
+
+      query =
```
bmquinn, line 339: Looks like you can yank this `query` variable being set here and just use `attrs` (I saw an unused variable warning in the console)

bmquinn (changes requested):

test/meadow/batches_test.exs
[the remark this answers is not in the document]
kdid: I don't think there needs to be an actual link between the tables. We just need to index the id's so we can link into a faceted search. 

I hear what you are saying but I think we already have had discussions about how our postgres record of the work is not the preservation object and it has some other stuff on it - like the ingest_sheet_id. I remember talking about that in regards to ingest_sheets and projects....

I was just trying to do the simplest thing possible. We can talk further tomorrow if you want.

mbklein:
I got this error while trying to submit a batch.

image

I think `add`, `delete`, `replace`, and `query` might need to be of type `text`, not `string(255)`.

bmquinn (approved):
Looks awesome, my only suggestion would be that we should keep an eye on reducing the 6 minute interval. My guess is we can lower it since the GenServer behaviour does a nice job of handling long-running interval tasks, which would mean less wait time for metadata folks 😄

mbklein (approved):
Agree with @bmquinn. The interval is the time between the _end_ of one run and the _beginning_ of the next, so a new task will never be fired while an existing one is running (assuming there's only one process running them). But I don't think that's a reason to hold up merging.

PR #1522, opened 2020-11-23
Increase time limits to avoid timeouts on large TIFFs

mbklein:
The pipeline has two different time limits built into it:

1. The `CreatePyramidTiff` action has a [timeout value](https://github.com/nulib/meadow/blob/1207-pyramid-tiff-timeout/lib/meadow/pipeline/actions/create_pyramid_tiff.ex#L11) that causes the operation to fail if Elixir hasn't heard back from the JavaScript port in a certain amount of time.
2. Each queue has a [`visibility_timeout` value](https://github.com/nulib/meadow/blob/1207-pyramid-tiff-timeout/config/sequins.exs#L24-L40) (default: 30 seconds). Exceeding the timeout doesn't cause a failure, but does put the message back into the queue for reprocessing if it hasn't been acknowledged after a certain amount of time. 

Setting the `visibility_timeout` too low can lead to a FileSet being processed by the same action (and all dependent actions) more than once. Since our actions are idempotent, this doesn't cause data problems, but it's inefficient.
Setting the `visibility_timeout` too high means that if an action really _does_ get stuck or fail in a way that Sequins and Broadway don't recognize, it will take longer for that message/action to retry.

Since a lot of this is (educated) guesswork, I'd argue for erring on the side of “too high.”

This PR:

* Increases the `visibility_timeout` on all actions that copy/stream TIFFs (digest/copy/derive) from 30 seconds to 5 minutes
* Adds a pingback message to the `CreatePyramidTiff` JavaScript code to avoid timeouts while the TIFF downloads from S3

### DIFF
--- a/config/sequins.exs
+++ b/config/sequins.exs
@@ -24,15 +24,15 @@ config :sequins, Meadow.Pipeline,
 config :sequins, IngestFileSet, queue_config: [processor_concurrency: 1]
 
 config :sequins, GenerateFileSetDigests,
-  queue_config: [max_number_of_messages: 3, visibility_timeout: 180],
+  queue_config: [max_number_of_messages: 3, visibility_timeout: 300],
   notify_on: [IngestFileSet: [status: :ok], GenerateFileSetDigests: [status: :retry]]
 
 config :sequins, CopyFileToPreservation,
-  queue_config: [max_number_of_messages: 3, visibility_timeout: 180],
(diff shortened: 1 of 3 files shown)


kdid:
Looks good to me.

kdid (approved):

bmquinn (approved):
Looks good, tested on a 2.43GB tif:
big_tif

PR #1526, opened 2020-11-24
Access control

divyareddynu:
Multiple roles and permissions have been added to `assets/js/components/Auth/DisplayAuthorized.jsx` component. The component handles display of buttons and other components based on user roles.

Closes [1161](https://github.com/nulib/repodev_planning_and_docs/issues/1161)


kdid (approved):
I'm not really following the logic of how this is working in the code, but it seems to work as expected in the UI.

adamjarling (approved):
Looks good.  @kdid This implements a "higher order component/function" pattern, for abstracting how we show/hide UI elements.  Maybe @divyareddynu could give a show & tell after standup?

PR #1542, opened 2020-12-01
Implement metadata update CSV export

mbklein:
## Testing

There is no GraphQL API for this yet because it was out of scope for the ticket. To test from iex:

```
iex> query = ~s({"query": {"term": {"model.name.keyword": "Image"}}})
iex> File.write("test.csv", Meadow.Data.CsvExport.generate_csv(query))
```

Then look at `test.csv` and make sure it looks right. Obviously, you can test with other queries as well, but things might get confusing if the result set contains non-image (e.g., FileSet or Collection) records.

### DIFF
--- a/lib/meadow/data/csv_export.ex
+++ b/lib/meadow/data/csv_export.ex
@@ -1,0 +1,142 @@
+defmodule Meadow.Data.CSVExport do
+  @moduledoc """
+  Generates a CSV representation of works matching an Elasticsearch query
+  """
+
+  alias Meadow.Data.Schemas.{WorkAdministrativeMetadata, WorkDescriptiveMetadata}
+  alias Meadow.Utils.ElasticsearchResultStream
+  alias NimbleCSV.RFC4180, as: CSV
+
+  @exportable_types ["Image"]
+  @top_level_fields [
+    ["id"],
+    ["accessionNumber"],
+    ["collection", "id"],
+    ["published"],
+    ["visibility"]
+  ]
+
+  @doc """
+  Generates the CSV output for an Elasticsearch query
+
+  Example:
+    iex> generate_csv(%{query: %{match_all: %{}}})
+    "{\"query\":{\"match_all\":{}}}\r\nid,accession_number..."
+  """
+  def generate_csv(query) when is_binary(query) do
+    query
+    |> stream_csv()
+    |> Enum.join("")
+  end
+
+  def generate_csv(query), do: Jason.encode!(query) |> generate_csv()
+
+  def stream_csv(query) when is_binary(query) do
+    Stream.resource(
+      fn -> :header end,
+      fn
+        nil -> {:halt, nil}
+        :header -> {generate_header(query), :rows}
+        :rows -> {generate_rows(works_only_query(query)), nil}
+      end,
+      fn _ -> :ok end
+    )
+    |> Stream.map(fn thing -> IO.iodata_to_binary(thing) end)
+  end
+
+  def stream_csv(query), do: Jason.encode!(query) |> stream_csv()
+
+  defp works_only_query(query) do
+    with actual_query <- Jason.decode!(query) |> Map.get("query") do
+      %{
+        query: %{
+          bool: %{
+            must: [
+              %{terms: %{"model.name.keyword" => @exportable_types}},
+              actual_query
+            ]
+          }
+        }
+      }
+      |> Jason.encode!()
+    end
+  end
+
+  defp fields do
+    @top_level_fields ++
+      fields_for(WorkAdministrativeMetadata, "administrativeMetadata") ++
+      fields_for(WorkDescriptiveMetadata, "descriptiveMetadata")
+  end
+
+  defp fields_for(module, prefix) do
+    apply(module, :field_names, [])
+    |> Enum.map(fn field -> [prefix, Inflex.camelize(field, :lower)] end)
+  end
+
+  defp generate_header(query) do
+    with fields <- fields() do
+      [
+        [query | List.duplicate(nil, length(fields) - 1)],
+        fields
+        |> Enum.map(fn field_path ->
+          case field_path |> List.last() do
+            "id" -> field_path |> Enum.map(&Inflex.underscore/1) |> Enum.join("_")
+            field_name -> Inflex.underscore(field_name)
+          end
+        end)
+      ]
+      |> CSV.dump_to_stream()
+    end
+  end
+
+  defp generate_rows(query) do
+    query
+    |> ElasticsearchResultStream.results()
+    |> Stream.map(&to_row/1)
+    |> CSV.dump_to_stream()
+  end
+
+  defp to_row(hit) do
+    fields()
+    |> Enum.map(fn field_path ->
+      hit
+      |> get_in(["_source" | field_path])
+      |> to_field()
+    end)
+  end
+
+  defp to_field(value) when is_list(value) do
+    value
+    |> Enum.map(&to_field/1)
+    |> Enum.join(" | ")
+  end
+
+  defp to_field(%{"edtf" => value}), do: value
+
+  defp to_field(%{"role" => role, "term" => %{"id" => id}}) do
+    case normalize_coded_term(role) do
+      nil -> id
+      prefix -> [prefix, id] |> Enum.join(":")
+    end
+  end
+
+  defp to_field(%{"label" => related_url_label, "url" => url}) do
+    case normalize_coded_term(related_url_label) do
+      nil -> url
+      prefix -> [prefix, url] |> Enum.join(":")
+    end
+  end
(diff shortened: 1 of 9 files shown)


kdid:
@mbklein - like the approach. Can you please connect this PR to an issue?

I'm seeing some issues with a couple fields:

```
** (Protocol.UndefinedError) protocol String.Chars not implemented for %{"id" => "UNIVERSITY_MAIN_LIBRARY", "label" => "University (MAIN) Library", "scheme" => "library_unit"} of type Map. This protocol is implemented for the following type(s): Phoenix.LiveComponent.CID, Postgrex.Copy, Postgrex.Query, Decimal, Float, DateTime, Time, List, Version.Requirement, Atom, Integer, Version, Date, BitString, NaiveDateTime, URI
    (elixir 1.10.4) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir 1.10.4) lib/string/chars.ex:22: String.Chars.to_string/1
    (elixir 1.10.4) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (elixir 1.10.4) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (elixir 1.10.4) lib/stream.ex:572: anonymous fn/4 in Stream.map/2
    (elixir 1.10.4) lib/enum.ex:3686: Enumerable.List.reduce/3
    (elixir 1.10.4) lib/stream.ex:1462: Stream.do_list_resource/6
    (elixir 1.10.4) lib/stream.ex:1609: Enumerable.Stream.do_each/4
    (elixir 1.10.4) lib/enum.ex:3383: Enum.map/2
    (meadow 0.1.0) lib/meadow/data/csv_export.ex:18: Meadow.Data.CsvExport.generate_csv/1
```

mbklein:
Nice find. I forgot to handle coded terms at the top level. Should be fixed now (and included in the test data).

lib/meadow/data/csv_export.ex
```diff
@@ -0,0 +1,86 @@
...
+  defp to_field(value) when is_list(value) do
+    value
+    |> Enum.map(&to_field/1)
+    |> Enum.join(" | ")
+  end
+
+  defp to_field(%{"role" => role, "term" => %{"label" => label}}) do
+    case normalize_role(role) do
+      nil -> label
+      prefix -> [prefix, label] |> Enum.join(":")
```
davidschober, line 72: @mbklein I know this is a bit ahead of scope, but if this is the human readable csv, we should format like "Klein, Michael (Developer)" rather than "dev:Klein, Michael". A use case will be to use this data in other systems, and I don't think the prefixes will make much sense. As I'm writing this, maybe it makes sense to have a future feature that allows a "format" flag.

mbklein: Oh, OK. Cool. That makes a lot of sense.

kdid: Wait. @davidschober - I thought we weren't doing the human readable csv? We're just doing the export/reimport?

mbklein: @kdid is right. I implemented this incorrectly from the get-go. Fortunately, the code is flexible enough to make it an easy fix. 😄

mbklein:
This code is now correctly exporting IDs instead of labels for controlled and coded terms and using `role_id:term_id` format instead of human-readable `Term Label (Role Label)` format.

kdid:
Here's a few things I noticed:

- `published` not exported
-  collection_id not exported
-  `visibility` not exported 
- `date_created` throws error
- `related_url` throws error

```
** (Protocol.UndefinedError) protocol String.Chars not implemented for %{"edtf" => "~1999", "humanized" => "circa 1999?"} of type Map. This protocol is implemented for the following type(s): Phoenix.LiveComponent.CID, Postgrex.Copy, Postgrex.Query, Decimal, Float, DateTime, Time, List, Version.Requirement, Atom, Integer, Version, Date, BitString, NaiveDateTime, URI
    (elixir 1.10.4) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir 1.10.4) lib/string/chars.ex:22: String.Chars.to_string/1
    (elixir 1.10.4) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (meadow 0.1.0) lib/meadow/data/csv_export.ex:65: Meadow.Data.CsvExport.to_field/1
    (elixir 1.10.4) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (elixir 1.10.4) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (elixir 1.10.4) lib/stream.ex:572: anonymous fn/4 in Stream.map/2
    (elixir 1.10.4) lib/enum.ex:3686: Enumerable.List.reduce/3
    (elixir 1.10.4) lib/stream.ex:1462: Stream.do_list_resource/6
    (elixir 1.10.4) lib/stream.ex:1609: Enumerable.Stream.do_each/4
    (elixir 1.10.4) lib/enum.ex:3383: Enum.map/2
    (meadow 0.1.0) lib/meadow/data/csv_export.ex:20: Meadow.Data.CsvExport.generate_csv/1
```


```
** (Protocol.UndefinedError) protocol String.Chars not implemented for %{"label" => %{"id" => "HATHI_TRUST_DIGITAL_LIBRARY", "label" => "Hathi Trust Digital Library", "scheme" => "related_url"}, "url" => "http://wow.com"} of type Map. This protocol is implemented for the following type(s): Phoenix.LiveComponent.CID, Postgrex.Copy, Postgrex.Query, Decimal, Float, DateTime, Time, List, Version.Requirement, Atom, Integer, Version, Date, BitString, NaiveDateTime, URI
    (elixir 1.10.4) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir 1.10.4) lib/string/chars.ex:22: String.Chars.to_string/1
    (elixir 1.10.4) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (meadow 0.1.0) lib/meadow/data/csv_export.ex:65: Meadow.Data.CsvExport.to_field/1
    (elixir 1.10.4) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (elixir 1.10.4) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (elixir 1.10.4) lib/stream.ex:572: anonymous fn/4 in Stream.map/2
    (elixir 1.10.4) lib/enum.ex:3686: Enumerable.List.reduce/3
    (elixir 1.10.4) lib/stream.ex:1462: Stream.do_list_resource/6
    (elixir 1.10.4) lib/stream.ex:1609: Enumerable.Stream.do_each/4
    (elixir 1.10.4) lib/enum.ex:3383: Enum.map/2
    (meadow 0.1.0) lib/meadow/data/csv_export.ex:20: Meadow.Data.CsvExport.generate_csv/1
```

lib/meadow/data/csv_export.ex
```diff
@@ -0,0 +1,98 @@
+defmodule Meadow.Data.CsvExport do
```
bmquinn, line 1:
```suggestion
defmodule Meadow.Data.CSVExport do
```

Any objections to capitalizing "CSV"? Thinking about `NimbleCSV` and how we went for `Meadow.IIIF` as well

mbklein:
I went ahead and added a protected route at `POST /apl/export/[filename].csv`. Because `CSVExport` uses Elixir Streams all the way through and doesn't unfold it until it absolutely has to, the controlled can return a chunked HTML response. This means that instead of waiting while the server generates the whole CSV and then downloading it all at once, the browser will start receiving data immediately and stream the response as it's generated.

The downside is that this means the server can't provide a `Content-Length` up front, so the browser can't show an accurate progress bar. But I consider that a pretty small tradeoff for a near-immediate time-to-first-byte.

kdid:
I'm still not getting the collection exported even when works belong to collections.


(I'm seeing 20-40 seconds for ~8000 works with metadata)

mbklein:
Ah, yeah, the collection ID is indexed differently from everything else on the export list. Rats. OK, that'll get fixed first thing this morning.

I think 20-40 seconds – or even longer – is reasonable, but only because there will be no “dead time” waiting for the download to start no matter how big the export.

lib/meadow/utils/data_loader.ex
```diff
@@ -13,6 +13,8 @@ defmodule Meadow.Utils.DataLoader do
     WorkDescriptiveMetadata
```
bmquinn, line 13:
```suggestion
```

Small request, you removed the reference to `WorkDescriptiveMetadata` so you can yank it from the alias list.

bmquinn (changes requested):

kdid (approved):

PR #1543, opened 2020-12-02
Validate and approve ingest sheets automatically

bmquinn:
Refactors `MeadowWeb.Resolvers.create_ingest_sheet/3` to validate and approve valid ingest sheet uploads automatically. 

Some UI updates...

image

image

image

### DIFF
--- a/lib/meadow_web/resolvers/ingest.ex
+++ b/lib/meadow_web/resolvers/ingest.ex
@@ -64,13 +64,29 @@ defmodule MeadowWeb.Resolvers.Ingest do
     {:ok, Sheets.get_ingest_sheet!(id)}
   end
 
-  def approve_ingest_sheet(_, %{id: id}, _) do
-    id
-    |> Sheets.get_ingest_sheet!()
-    |> approve_ingest_sheet()
+  def ingest_sheet_validation_progress(_, %{id: id}, _) do
+    {:ok, Sheets.get_sheet_validation_progress([id]) |> Map.get(id)}
   end
 
-  def approve_ingest_sheet(%{status: "valid"} = ingest_sheet) do
+  def ingest_sheet_validations(_, _, _) do
+    {:ok, %{validations: [%{id: "sheet", object: %{errors: [], status: "pending"}}]}}
+  end
+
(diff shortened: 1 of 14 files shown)


lib/meadow_web/resolvers/ingest.ex
```diff
@@ -123,7 +123,18 @@ defmodule MeadowWeb.Resolvers.Ingest do
          details: ChangesetErrors.error_details(changeset)}
 
       {:ok, ingest_sheet} ->
-        {:ok, ingest_sheet}
+        Validator.validate(ingest_sheet.id)
```
mbklein, line 126: The only problem I see here is that the mutation won't return until validation completes and ingest starts. This might cause problems (or timeouts) for large ingest sheets.

mbklein (changes requested):

lib/meadow_web/resolvers/ingest.ex
bmquinn: Good call. Any suggestions on how to avoid those problems? I feel like that's going to be tricky to solve with how we do validation and approval currently.

mbklein: Maybe move the auto-approve code out of the resolver and into the validator itself? Maybe even pass an `auto_approve` parameter to the validator?

bmquinn: Thanks for the assist @mbklein and @kdid!

bmquinn:
@adamjarling this is ready you to take over. Everything works as you would expect when uploading an ingest sheet, you go straight to validation and approval. Probably mostly just cleanup of components related to manually approving a sheet. Just a heads up we removed the `approve_ingest_sheet` mutation completely since it is redundant now.

adamjarling:
@bmquinn Running this locally just to see what it does, I get the following ingesting a sheet.  Were you seeing this?  It might be unrelated...looks like it his the approval status successfully.    I'll try a few more times.

image

image

bmquinn:
@adamjarling I think MBK answered in Slack, but just in case running `mix meadow.pipeline.setup` will fix that error.

kdid (approved):

PR #1547, opened 2020-12-02
Clear input if selection not from controlled term list

divyareddynu:
This PR clears controlled term input on blur if the value is not from the list of controlled terms provided.


adamjarling (approved):
Looks great and works as expected

PR #1551, opened 2020-12-03
1183 Support sending an Elasticsearch query to the Search screen

divyareddynu:
Tried making Reactivesearch components automatically know about a `query` state which is passed to the `ResultsList` Reactivesearch component, but (to the extent I know how to try), it's not possible.

This uses a custom parser function which will extract any Search text or Facet values from a provided Elasticsearch query.

When dealing with individually selected items, it's a little wonky, and if we want to "remember" the actual search and / or facet values this will take some time to wire up.

image

image


kdid:
@adamjarling or @divyareddynu - So, when we say supports an Elasticsearch query, do we mean like copying and pasting one from a batch details page into the search box?

adamjarling:
@kdid No, it means that if an ES query is being passed around, either from a saved search, or a batch dashboard job, or from whereever, if it's passed into the Search screen component as a `state` variable, then the Search screen will pick it up and create the UI selections that created the query.

Make sense?

adamjarling:
Although the way we now have it, we could probably support a non-ReactiveSearch textarea component where a user could enter a query.

kdid:
So what should I see happening in the UI as a result of this PR? Or is this just behind the scenes?

kdid:
Ah I see, yes that works for me. Code looks fine to me. I'm approving though my 2 cents is that the "Back to saved search" button didn't register with me. I don't think I would have viewed this as a "saved search" from the UI perspective (saved searches I've done on other sites involved clicking some save buttons and naming things, etc). I think "back to search results" or something would have made more sense. 

But that's just me...

kdid (approved):

divyareddynu (commented):
This is sweet! Thank you for taking over this issue. I can't approve it as it's my own PR, but yes, this is perfect!

adamjarling:
Yea, we can label the button whatever... I'll updated it on my next issue just to get this one through for now.

PR #1553, opened 2020-12-03
Add sort options

divyareddynu:
Screen Shot 2020-12-03 at 3 47 55 PM

### DIFF
--- a/assets/js/services/global-vars.js
+++ b/assets/js/services/global-vars.js
@@ -3,7 +3,7 @@ export const COLLECTION_TYPES = [
   { label: "NUL Theme", id: "NUL Theme", value: "NUL Theme" },
 ];
 
-export const SORT_OPTIONS = [
+export const REACTIVESEARCH_SORT_OPTIONS = [
   {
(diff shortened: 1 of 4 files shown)


assets/js/services/global-vars.js
```diff
@@ -3,6 +3,24 @@ export const COLLECTION_TYPES = [
...
+    dataField: "modifiedDate",
+    label: "Sort By Modified Date",
+  },
+  {
+    sortBy: "asc",
+    dataField: "descriptiveMetadata.title.keyword",
+    label: "Sort By Title",
+  },
+];
+
```
adamjarling, line 23: @divyareddynu I know it's small, but can we make this variable constant title a little more specific?  How about something like `REACTIVESEARCH_SORT_OPTIONS`?   Think of doing quick scans like in 6 months of the code, and what will help us quickly understand it without even looking at the code.   Everything else here looks great.

adamjarling (changes requested):

adamjarling (approved):

PR #1561, opened 2020-12-07
Lock db when retrieving pending works in WorkCreator

kdid:
- Locks the db row after retrieving `pending` works until they are set to `processing` so that more than one process can't grab them.

### DIFF
--- a/lib/meadow/ingest/work_creator.ex
+++ b/lib/meadow/ingest/work_creator.ex
@@ -28,20 +28,32 @@ defmodule Meadow.Ingest.WorkCreator do
   """
   def create_works(state) do
     with_log_level(:info, fn ->
-      case Progress.get_pending_work_entries(state.batch_size)
-           |> Progress.update_entries("CreateWork", "processing")
-           |> Repo.preload(row: :sheet) do
-        [] ->
-          :noop
-
-        work_rows ->
-          Logger.enable(self())
-          Logger.info("Creating #{length(work_rows)} works")
-          create_works_from_rows(work_rows)
-      end
+      state
+      |> get_and_update_pending_work_rows()
+      |> handle_result()
     end)
 
     {:noreply, state}
+  end
+
+  defp get_and_update_pending_work_rows(state) do
+    Repo.transaction(fn ->
+      Progress.get_and_lock_pending_work_entries(state.batch_size)
+      |> Progress.update_entries("CreateWork", "processing")
+    end)
+  end
+
+  defp handle_result({:ok, []}), do: []
+
+  defp handle_result({:ok, progress_rows}) do
+    Logger.enable(self())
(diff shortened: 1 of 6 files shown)


mbklein (approved):

lib/meadow/ingest/work_creator.ex
```diff
@@ -23,21 +23,26 @@ defmodule Meadow.Ingest.WorkCreator do
     %{batch_size: Keyword.get(args, :batch_size, 20)}
   end
 
+  @spec create_works(any) :: {:noreply, any}
   @doc """
   Turn a batch of pending work rows into works
   """
   def create_works(state) do
```
bmquinn, line 30: I've got a refactoring suggestion if you'd indulge me 😄 . There's a lot going on in `create_works/1` that I think could benefit from a change like this:

```suggestion
  def create_works(state) do
    with_log_level(:info, fn ->
      state
      |> get_and_update_pending_work_rows()
      |> handle_result()
    end)

    {:noreply, state}
  end

  defp get_and_update_pending_work_rows(state) do
    Repo.transaction(fn ->
      Progress.get_and_lock_pending_work_entries(state.batch_size)
      |> Progress.update_entries("CreateWork", "processing")
    end)
  end

  defp handle_result({:ok, []}), do: :noop

  defp handle_result({:ok, progress_rows}) do
    Logger.enable(self())
    Logger.info("Creating #{length(progress_rows)} works")
    create_works_from_rows(progress_rows |> Repo.preload(row: :sheet))
  end

  defp handle_result({:error, message}),
    do: Logger.error("Problem processing work entries: #{message}")
```

The test still pass, but I threw the logging error in the last `handle_result/1` without a ton of consideration. What do you think?

bmquinn (changes requested):

lib/meadow/ingest/work_creator.ex
mbklein: I think this is a good refactoring, only I would return `[]` from the last `handle_result/1` just to respect the contract.

kdid:
@bmquinn and @mbklein - I made the updates. PTAL.

lib/meadow/ingest/work_creator.ex
```diff
@@ -28,22 +28,34 @@ defmodule Meadow.Ingest.WorkCreator do
...
   end
 
+  defp get_and_update_pending_work_rows(state) do
+    Repo.transaction(fn ->
+      Progress.get_and_lock_pending_work_entries(state.batch_size)
+      |> Progress.update_entries("CreateWork", "processing")
+    end)
+  end
+
+  defp handle_result({:ok, []}), do: :noop
```
bmquinn, line 46:
```suggestion
  defp handle_result({:ok, []}), do: []
```

kdid:
updated

bmquinn (approved):

PR #1562, opened 2020-12-07
Updated ingest sheet list

divyareddynu:
Screen Shot 2020-12-07 at 8 58 09 AM

### DIFF
--- a/assets/js/components/IngestSheet/List.jsx
+++ b/assets/js/components/IngestSheet/List.jsx
@@ -68,46 +68,49 @@ const IngestSheetList = ({ project, subscribeToIngestSheetStatusChanges }) => {
               </tr>
             </thead>
             <tbody>
-              {project.ingestSheets.map(({ id, title, status, updatedAt }) => (
-                <tr key={id}>
-                  <td>
-                    <Link to={`/project/${project.id}/ingest-sheet/${id}`}>
-                      {title}
-                    </Link>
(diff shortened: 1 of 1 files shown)


assets/js/components/IngestSheet/List.jsx
```diff
@@ -68,46 +68,49 @@ const IngestSheetList = ({ project, subscribeToIngestSheetStatusChanges }) => {
...
-                    {["APPROVED", "COMPLETED"].indexOf(status) > -1 && (
-                      <Link
-                        to={`/project/${project.id}/ingest-sheet/${id}`}
-                        className="button"
-                      >
-                        {<FontAwesomeIcon icon="eye" />}{" "}
-                        <span className="sr-only">View</span>
+              {project.ingestSheets.map(
+                ({ id, title, status, updatedAt }, index) => (
+                  <tr key={id + index}>
```
adamjarling, line 73: Is `id` not unique here?  How does adding the `index` value help?

divyareddynu: The actual bug is that there is a console error for duplicity of keys and then the rows wouldn't update once the error occurs. Id is not unique due to changing status messages (like validation pending/validated etc where id is same but status is not). Adding index provides uniqueness so there are no errors and the rows update when state changes.

adamjarling (approved):

PR #1563, opened 2020-12-07
Wireup fileset reorder query

divyareddynu:


adamjarling (changes requested):
Everything works great.  While we're in the code can we make a few quick UI enhancements?

On fileset reorder, can we update the success Toast message to say "Filesets have been successfully reordered."    

image

image

adamjarling (approved):

PR #1570, opened 2020-12-08
expose verify FileSet API

kdid:
@adamjarling - ended up implementing this so it takes a work id rather than just an array of file set ids. When I started working on it that seemed to make sense - that way you could make the request on page load rather having to make a second request after load by sending back the file set id's in an array. Let me know if that doesn't seem right to you.  

Screen Shot 2020-12-07 at 5 22 45 PM

### DIFF
--- a/lib/meadow/data/works.ex
+++ b/lib/meadow/data/works.ex
@@ -540,4 +540,24 @@ defmodule Meadow.Data.Works do
     |> Multi.update(:work, work |> Work.update_timestamp())
     |> Repo.transaction()
   end
+
+  def verify_file_sets(work_id) do
+    work_id
+    |> with_file_sets()
+    |> Map.get(:file_sets)
+    |> Enum.map(fn file_set ->
+      %{file_set_id: file_set.id, verified: verify_file_set(file_set)}
+    end)
+  end
+
+  defp verify_file_set(file_set) do
+    case ExAws.S3.head_object(
+           Config.preservation_bucket(),
+           URI.parse(file_set.metadata.location).path
+         )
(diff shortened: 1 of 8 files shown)


adamjarling:
@kdid That makes sense.  Whenever in doubt, I'd say harness the dynamic nature of GraphQL and minimize network requests for sure.

adamjarling (approved):
This response shape looks good to handle in the UI.

lib/meadow/data/works.ex
```diff
@@ -540,4 +540,24 @@ defmodule Meadow.Data.Works do
...
     |> Repo.transaction()
   end
+
+  def verify_file_sets(work_id) do
+    work = with_file_sets(work_id)
+
+    work.file_sets
+    |> Enum.map(fn file_set ->
+      %{file_set_id: file_set.id, verified: verify_file_set(file_set)}
+    end)
```
bmquinn, line 550: You can skip the variable assignment here with the ol' pipe operator 😄 :

```suggestion
    work_id
    |> with_file_sets()
    |> Enum.map(fn file_set ->
      %{file_set_id: file_set.id, verified: verify_file_set(file_set)}
    end)
```

bmquinn (changes requested):
Looks awesome, I really like the id of passing `workId` for simplicity. Just a small change request for `verify_file_sets/1` otherwise a big 👍 from me.

lib/meadow/data/works.ex
kdid: updated.

bmquinn (approved):

PR #1573, opened 2020-12-08
Wireup relatedUrl batchEdit

divyareddynu:
Related URL is now wired up and is set to post on batchEdit. Updates might a bit longer but it works.
Screen Shot 2020-12-08 at 2 19 08 PM


adamjarling (approved):
Looks good

PR #1578, opened 2020-12-09
Support preservation masters (backend)

kdid:
- updates `dc_download` task so that PM's are present in the generated ingest sheet
- Reconfigures pipeline so derivatives are not generated for preservation masters
- Updates the front end to only show access masters on structure tab and only allow access masters to participate in file set order.
- excludes PM's from IIIF manifest
- updated db trigger to rewrite manifest if rank is updated
- changes set_representative_image so it only sets image to an access master
- reconfigures `EctoRanked` for FileSets so it is scoped to `work_id` + `role` 
- created issue to refactor code associated with `Pipeline.actions( )` https://github.com/nulib/repodev_planning_and_docs/issues/1323

also fixes https://github.com/nulib/repodev_planning_and_docs/issues/1315

To merger: 
- Please let Divya and Adam know to run a `devstack down -v`
- Please reset staging

### DIFF
--- a/assets/js/components/Work/Tabs/Structure/Structure.jsx
+++ b/assets/js/components/Work/Tabs/Structure/Structure.jsx
@@ -7,7 +7,7 @@ import {
   GET_WORK,
   SET_WORK_IMAGE,
   UPDATE_FILE_SETS,
-  UPDATE_FILE_SET_ORDER,
+  UPDATE_ACCESS_MASTER_ORDER,
 } from "@js/components/Work/work.gql.js";
 import { toastWrapper } from "@js/services/helpers";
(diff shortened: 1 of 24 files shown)


adamjarling (approved):
Stuff looks good on the front-end.  Is there ever a need for the UI to know about, or handle preservation master files?

kdid:
@adamjarling they are on the Preservation tab only.

bmquinn (approved):
Looks great @kdid! Glad you added the role randomization to `dc_download` so there's one less thing to think about when seeding test data too.

mbklein (approved):
This is great. Merging to staging will require a little more than a `reset`, because we have to delete the existing SQS/SNS subscriptions first, but otherwise it should be pretty smooth.

PR #1579, opened 2020-12-09
1259 batch dashboard

adamjarling:
Batch Dashboard and Details screens added.

image

image


kdid:
Looks great. The only thing I noticed is that when the batch is in the queued state, started says "Invalid Date". Maybe just blank or "--" or something would be better...

It looks like you did this with polling? No subscription? Code looks clean and certainly user experience same or better....


invalid-date-batch

adamjarling:
@kdid Got it.  I'll fix the date display issue.

Regarding polling, yea seems based on this use case and probably others we've had in the past, Apollo recommends it:  https://www.apollographql.com/docs/react/data/queries/#polling

If it results in less code to maintain and write on the backend, seems for the data we're passing around we may not need subscriptions at all.

kdid:
Yeah. I'm going to write up an issue to get rid of the backend subscription code. No need to keep it around.

kdid (approved):

PR #1586, opened 2020-12-10
Move Delete to Preservation Tab

divyareddynu:
Screen Shot 2020-12-10 at 11 01 12 AM

### DIFF
--- a/assets/js/components/Work/Tabs/Preservation.jsx
+++ b/assets/js/components/Work/Tabs/Preservation.jsx
@@ -90,7 +90,7 @@ const WorkTabsPreservation = ({ work }) => {
       </div>
       <div className="container buttons">
         <DisplayAuthorized action="delete">
-          <Button data-testid="delete-button" onClick={onOpenModal}>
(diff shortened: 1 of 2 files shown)


adamjarling:
I think there may be some confusion here... in that the "Delete Work" button got moved to the bottom of the Work screen a while ago (probably right after usability testing), and is independent of any of the tabs on the Work screen.

@davidschober Check out this screenshot... or view on Meadow staging for how this currently works.   If this is ok as-is, I'll just close the PR, but if you want the "Delete Work" action (not to be confused with deleting a Fileset, which the Preservation tab currently supports), to only show up on the Preservation tab, then I'll continue reviewing this PR.

mix meadow.seed_data wwii_posters
image

davidschober:
I think it should only be on the preservation tab. The idea is that this is a serious thing and it is a preservation action. Maybe we can pop an "are you sure" window. We should also make sure that this is managers or above.

adamjarling (approved):

PR #1604, opened 2020-12-14
Switch default search bar operator to AND instead of OR

adamjarling:
Now searches in the search bar will use `AND` instead of `OR` on multi word searches.

image


mbklein (approved):

davidschober:
This is interesting. What search type are we using I thought we were using simple query string query, but this seems to be respecting the words OR and AND rather than the symbols + -. I'm good either way, but I'm a little confused as it's different than DC.

adamjarling:
@davidschober I can walk you through the setup in Meadow if you want to see how ReactiveSearch is configured there, and what the ReactiveSearch api says.

davidschober:
Cool. That will be helpful

PR #1615, opened 2020-12-17
Adds ability to add fileset to work in the UI

kdid:
- Adds ability to add FileSet to a work through the Preservation tab.ts.
  - refactors `Meadow.Ingest.Bucket` functions and `getPresignedUrl` mutation to be more generic and deal with presigned url's for ingest sheets and file sets
  - adds `ingestFileSet` mutation to GraphQL + related Elixir/Phoenix code
  - wires up front end form

Screen Shot 2020-12-18 at 8 01 21 AM

Opened issue for manifest refresh (for IIIF viewer) after save: https://github.com/nulib/repodev_planning_and_docs/issues/1346

Other issues opened for future enhancements - potentially showing fileset progress through ingest pipeline when not part of ingest sheet:
https://github.com/nulib/repodev_planning_and_docs/issues/1344


kdid:
@adamjarling - after I rebased against your preservation tab branch the Add file set button is getting chopped. Any ideas why?

Screen Shot 2020-12-18 at 7 39 52 AM

assets/js/client.js
```diff
@@ -39,7 +39,15 @@ const link = new ApolloLink.split(
...
-  cache: new InMemoryCache(),
+  cache: new InMemoryCache({
+    typePolicies: {
+      FileSet: {
+        fields: {
+          metadata: { merge: false },
+        },
+      },
+    },
+  }),
```
adamjarling, line 50: Why is this needed?

assets/js/components/Work/Tabs/FileSetModal.jsx
```diff
@@ -0,0 +1,309 @@
...
+import UIFormSelect from "@js/components/UI/Form/Select.jsx";
+import { FILE_SET_ROLES } from "@js/services/global-vars";
+import UISkeleton from "@js/components/UI/Skeleton";
+
+/** @jsx jsx */
+import { css, jsx } from "@emotion/core";
+const dropZone = css`
+  background: #efefef;
+  border: 3px dashed #ccc;
+`;
```
adamjarling, line 23: We should probably globalize these styles into `.scss` now that we're re-using the React Dropzone component.

assets/js/components/Work/Tabs/FileSetModal.jsx
```diff
@@ -0,0 +1,309 @@
...
+  };
+
+  const onSubmit = async (data) => {
+    const requestOptions = {
+      method: "PUT",
+      headers: { "Content-Type": "multipart/form-data" },
+      body: currentFile,
+    };
+
+    fetch(urlData.presignedUrl.url, requestOptions)
```
adamjarling, line 89: Nice replacement for `axios`

assets/js/components/Work/Tabs/FileSetModal.jsx
```diff
@@ -0,0 +1,309 @@
...
+          },
+        });
+      })
+      .catch((error) => {
+        console.error("There was a problem with the S3 upload", error);
+        toastWrapper(
+          "is-danger",
+          `There was a problem with the S3 upload: ${error}`
+        );
+      });
```
adamjarling, line 117: I'm curious here how an error in the `ingestFileSet` mutation would be handled.  Does the `fetch .catch()` block catch it?   Does it pass through to the mutation error handler?   

If it works, it works, but if there was any overlap in error handling that's a head scratcher, another option may be to use `async/await` to synchronously grab the presigned url, then call `ingestFileSet` and let Apollo client handle any errors.

assets/js/components/Work/Tabs/FileSetModal.jsx
```diff
@@ -0,0 +1,309 @@
...
+                    <UIFormField label="Accession number">
+                      <UIFormInput
+                        isReactHookForm
+                        required
+                        label="FileSet label"
+                        data-testid="fileset-accession-number-input"
+                        name="accession_number"
+                        placeholder="accession number"
+                        onChange={handleInputChange}
+                        value={values.accession_number}
```
adamjarling, line 185: If we really need to set the `value` here, then I'd argue we probably don't need the React Hook Form baggage.  Any way to get this to work without controlling the input's value?    I could try to mess with it if you want.

assets/js/components/Work/Tabs/FileSetModal.jsx
```diff
@@ -0,0 +1,309 @@
...
+                          </tbody>
+                        </table>
+                      </React.Fragment>
+                    )}
+                  </div>
+                )}
+              </section>
+
+              <footer className="modal-card-foot is-justify-content-flex-end">
+                {isSubmitting ? (
```
adamjarling, line 275: Instead of a ternary here, you could do a:

```
{ isSubmitting && (
...
  <Button isText .../>
)
}
```

assets/js/components/Work/Tabs/Preservation.jsx
```diff
@@ -75,6 +77,20 @@ const WorkTabsPreservation = ({ work }) => {
...
+        <div className="buttons is-right">
+          <Button
+            data-testid="button-new-file-set"
+            isPrimary
+            onClick={() => setIsModalHidden(!isModalHidden)}
+          >
+            <span className="icon">
+              <FontAwesomeIcon icon="file-image" />
+            </span>
+            <span>Add a FileSet</span>
```
adamjarling, line 90: On button labels, for consistency (and this isn't like documented anywhere), but we've been using a convention to only have the first word capitalized.  One day we should clean it up, but it's probably 50/50 on whether "file set" is one word, camelCased, all that fun stuff.

assets/js/services/helpers.js
```diff
@@ -111,3 +111,7 @@ export function prepWorkItemForDisplay(res) {
     workType: res.workType,
   };
 }
+
+export function s3Location(presignedUrl) {
+  return `s3://${presignedUrl.split("?")[0].split("/").slice(-3).join("/")}`;
+}
```
adamjarling, line 117: Nice

adamjarling:
@kdid I think the tab button getting cut off is because the button is outside the `<header>`.  If you tuck the columns div inside the header like the other tabs, it should render out evenly.

image

adamjarling:
I'd say we also get rid of the extra "Box" inside the modal itself.

image

assets/js/client.js
kdid: I was getting a warning about updating the cache because FileSetMetadata did not have id's.  

So followed this doc:
https://www.apollographql.com/docs/react/caching/cache-field-behavior/#merging-non-normalized-objects

assets/js/components/Work/Tabs/FileSetModal.jsx
kdid: If you're running the branch locally you can cause an error in the `ingestFileSet` mutation by trying to submit the form using an accession_number that already exists. The fetch `catch` block does not catch it. The `onError({ error })` in the mutation recognizes it  - though (and this was a head scratcher for me - the `error` object on line 57 is undefined but the error details are present in the `error` object and printed on line 166/168.  

I can try to rework it to wait for the presigned url on submit like you said.

assets/js/components/Work/Tabs/FileSetModal.jsx
kdid: I struggled with that. I couldn't get it to retain the form values after a submit with error. (like duplicate accession number) unless I manually set the values. 

If you'd like to try - great.

lib/meadow_web/resolvers/helpers.ex
```diff
@@ -4,6 +4,13 @@ defmodule MeadowWeb.Resolvers.Helpers do
   """
 
   alias Meadow.Config
+  alias Meadow.Utils.AWS
+
+  def get_presigned_url(_, params, _) do
+    with url <- AWS.presigned_url(Config.upload_bucket(), params) do
+      {:ok, %{url: url}}
+    end
```
bmquinn, line 12: Would it make sense to handle the error case here as well? Thinking something like:
```suggestion
    case AWS.presigned_url(Config.upload_bucket(), params) do
      {:ok, url} -> {:ok, url}
      {:error, msg} -> {:error, message: "Could not create presigned url", details: msg}
    end
```

lib/meadow/data/file_sets.ex
```diff
@@ -99,6 +100,17 @@ defmodule Meadow.Data.FileSets do
...
+    case create_file_set(attrs) do
+      {:ok, file_set} ->
+        Pipeline.kickoff(file_set, %{role: file_set.role})
+        {:ok, file_set}
+
+      {:error, changeset} ->
+        {:error, changeset}
+    end
+  end
+
```
bmquinn, line 113: I think we should consider moving this function out of the `FileSets` context. In my understanding, this module is mainly responsible for interacting with the Repo and `FileSet` schema. As another example, the `Works` context doesn't have any references to the `Pipeline` either. Could possibly be moved to `Meadow.Pipeline` or maybe even a new module?

bmquinn (commented):
Looks awesome. Had a couple questions, but don't let me hold you up while I'm out for the holidays.

lib/meadow_web/resolvers/helpers.ex
mbklein: I get the impulse, but there is a 0% chance that the call to `presigned_url` will fail without something else having failed earlier. It's a completely client-side operation that takes some data and builds a URL.

adamjarling (approved):
Looks good.  I vote we get it in.

mbklein (approved):

PR #1634, opened 2020-12-23
Refactor FileSetModal

kdid:
Refactors the FileSetModal
- breaks out into components
- Changes the flow:
   - uploads the file (with progress bar) immediately when dropped
   - hides the drop zone once a file has been dropped
   - shows the input fields once the upload is successful
   - handles aborting the upload (if someone closes the modal with the upload in progress)
   - displays GraphQL errors and file Upload errors (as well as an unlikely presigned url error)

Screen Shot 2020-12-23 at 11 19 35 AM

Screen Shot 2020-12-23 at 11 18 20 AM

### DIFF
--- a/assets/js/components/Work/Tabs/Preservation/FileSetModal.jsx
+++ b/assets/js/components/Work/Tabs/Preservation/FileSetModal.jsx
@@ -6,17 +6,16 @@ import { GET_WORK, INGEST_FILE_SET } from "@js/components/Work/work.gql.js";
 import { useQuery, useMutation } from "@apollo/client";
 import { s3Location, toastWrapper } from "@js/services/helpers";
 import { useForm, FormProvider } from "react-hook-form";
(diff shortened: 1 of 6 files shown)


assets/js/components/Work/Tabs/Preservation/FileSetModal.jsx
```diff
@@ -48,179 +45,155 @@ function FileSetModal({ closeModal, isHidden, workId }) {
...
+      uploadFile(file);
+    } else {
+      setUploadProgress(0);
+    }
   };
 
-  const clearErrors = () => {
-    setFileUploadError("");
-  };
+  const uploadFile = (file) => {
```
adamjarling, line 106: This looks great.  I think in the future we could abstract this out into its own component, or a custom hook... `useFileUploader` or something.

adamjarling (approved):
Great work.  I'd say get rid of the `console.log`s hanging around and then a nice improvement to the UX and UI.

PR #1660, opened 2021-01-08
Adjust image size for workImage

divyareddynu:
Screen Shot 2021-01-08 at 11 52 20 AM
Screen Shot 2021-01-08 at 11 52 33 AM


assets/js/components/UI/WorkImage.jsx
```diff
@@ -6,7 +6,13 @@ function UIWorkImage({ imageUrl = "", size = 128 }) {
     ? `${imageUrl}/square/${size},${size}/0/default.jpg`
     : `/images/480x480.png`;
 
-  return <img data-testid="image-source" src={src} />;
+  return (
+    <img
+      data-testid="image-source"
+      src={src}
+      className={`image is-${size}x${size}`}
```
adamjarling, line 13: Can you check this?  I think the Bulma image size classes are meant to go on a `<figure>` and not an `` tag.   And also I think we need a better placeholder image.  I'll send you one.

https://bulma.io/documentation/elements/image/

adamjarling:
How about this one:

image

adamjarling:
Nice, new changes look good.  Looks like the tests got knocked loose.

adamjarling (approved):

PR #1672, opened 2021-01-12
Update navbar z-index to display on top of OSD viewer

adamjarling:
Quick z-index update.

image


divyareddynu:
Still seeing this
Screen Shot 2021-01-12 at 1 30 43 PM

adamjarling:
Ah, good catch.  Didn't even think about the modals.  I'll fix and re-push.

divyareddynu (approved):

kdid (approved):

PR #1684, opened 2021-01-13
1397 wireup project editing

divyareddynu:
Screen Shot 2021-01-13 at 8 39 39 AM

### DIFF
--- a/assets/js/components/Project/Form.jsx
+++ b/assets/js/components/Project/Form.jsx
@@ -1,6 +1,6 @@
 import React, { useState } from "react";
 import { useMutation } from "@apollo/client";
-import { CREATE_PROJECT, GET_PROJECTS } from "./project.gql.js";
+import { CREATE_PROJECT, GET_PROJECTS, UPDATE_PROJECT } from "./project.gql.js";
 import { useForm, FormProvider } from "react-hook-form";
 import Error from "../UI/Error";
 import Loading from "../UI/Loading";
@@ -9,7 +9,7 @@ import UIFormInput from "../UI/Form/Input.jsx";
 import UIFormField from "../UI/Form/Field.jsx";
 import { Button } from "@nulib/admin-react-components";
 
-const ProjectForm = ({ showForm, setShowForm }) => {
+const ProjectForm = ({ showForm, setShowForm, project = {}, formType }) => {
   const [formError, setFormError] = useState();
   const methods = useForm();
   let [createProject, { loading, error: mutationError, data }] = useMutation(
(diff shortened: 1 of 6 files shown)


assets/js/components/Project/Form.jsx
```diff
@@ -31,12 +31,41 @@ const ProjectForm = ({ showForm, setShowForm }) => {
     }
   );
 
+  let [
+    updateProject,
+    {
+      loading: projectUpdateLoading,
+      error: projectUpdateError,
+      projectUpdateData,
```
adamjarling, line 39: I'd say if we're not using these values, let's not include the object argument and keep it cleaner.

assets/js/components/Project/Form.jsx
```diff
@@ -31,12 +31,41 @@ const ProjectForm = ({ showForm, setShowForm }) => {
...
+        `Project ${updateProject.title} updated successfully`
+      );
+      setShowForm(false);
+    },
+    onError(error) {
+      setFormError(error);
+    },
+    refetchQueries(mutationResult) {
+      return [{ query: GET_PROJECTS }];
+    },
```
adamjarling, line 54: Pretty sure as long as the `query` returns and item's `id`, we don't need to refetch the Queries.  It should update the Apollo cache automagically.

assets/js/components/Project/Form.jsx
```diff
@@ -58,12 +87,15 @@ const ProjectForm = ({ showForm, setShowForm }) => {
 
                 <UIFormField label="Project Title">
                   <UIFormInput
+                    id={project.id || formType}
```
adamjarling, line 90: Seems like we're using `id` for something other than the attribute is intended by putting `formType` value on it.  Is there another way to communicate this `formType`?

assets/js/components/Project/List.jsx
```diff
@@ -49,9 +52,15 @@ const ProjectList = () => {
...
 
+  const onEditProject = (e, project) => {
+    setActiveProject(project);
+    setShowForm(!showForm);
+  };
+
   const onCloseModal = () => {
     setActiveModal(null);
     setModalOpen(false);
+    setActiveProject();
```
adamjarling, line 63: Let's feed `setActiveProject()` a value.  Either `null` or an empty string `''` if it's a string data type.

assets/js/components/Project/List.jsx
```diff
@@ -49,9 +52,15 @@ const ProjectList = () => {
     setModalOpen(true);
   };
 
+  const onEditProject = (e, project) => {
```
adamjarling, line 55: Are we doing anything with `e` here?  If not, let's not pass it in.

adamjarling (changes requested):
A few random thoughts on this PR to clean up the code.  Functionality is great.

assets/js/components/Project/Form.jsx
divyareddynu: I tried with key, key only updates the value part. if the id is not unique, since we use the same form twice on a single page /projects to add and edit, the id is coming out as same causing this error 
Screen Shot 2021-01-20 at 9 51 29 AM

divyareddynu: I can do ` id={`${formType}${project.id}`}` to incorporate both and not make it awkward.

adamjarling (approved):
Looks good!

PR #1702, opened 2021-01-19
Add and configure exif lambda and ExtractExifMetadata pipeline action

bmquinn:
- Add `exif` map to `FileSetMetadata` embedded schema
- Add `exif` lambda that is called with the new lambda handler
- Add and configure `ExtractExifMetadata` pipeline action with tests
- Reduce lambda utility debug logging noise

Note: requires a `mix meadow.setup` to set up the new pipeline action

facets
exif_graphql


config/releases.exs
```diff
@@ -110,7 +110,13 @@ config :hackney,
...
+config :meadow, :lambda,
+  digester: {:lambda, "meadow-digester"},
+  exif: {:lambda, "meadow-exif"}
 
 config :sequins, Actions.GenerateFileSetDigests,
   queue_config: [processor_concurrency: 50, visibility_timeout: 300]
+
+config :sequins, Actions.ExtractExifMetadata,
+  queue_config: [processor_concurrency: 50, visibility_timeout: 300],
+  notify_on: [CopyFileToPreservation: [status: :ok], ExtractExifMetadata: [status: :retry]]
```
bmquinn, line 122: @mbklein should I skip the `notify_on` parameter here as in `Actions.GenerateFileSetDigests` above?

terraform/lambdas.tf
```diff
@@ -85,3 +85,17 @@ module "digester_function" {
...
 }
 
+module "exif_function" {
+  depends_on    = [ aws_iam_role_policy_attachment.lambda_bucket_access ]
+  source        = "./modules/meadow_lambda"
+  name          = "exif"
+  description   = "Function to extract EXIF metadata from an S3 object"
+  role          = aws_iam_role.lambda_role.arn
+  stack_name    = var.stack_name
+  memory_size   = 1024
```
bmquinn, line 95: @mbklein I copied the `memory_size` value from the digester function, but should we bump this up for processing large TIFFs?

assets/js/services/reactive-search.js
```diff
@@ -163,8 +163,114 @@ export const FACET_SENSORS = [
...
+//     ...defaultRangeItemValues,
+//     componentId: "YResolution",
+//     dataField: "fileSets.exif.model.yResolution",
+//     title: "YResolution",
+//     range: { start: 0, end: 1000 },
+//     rangeLabels: { start: "0px", end: "1000px" },
+//     defaultValue: { start: 0, end: 1000 },
+//   },
+// ];
+
```
adamjarling, line 273: Yea, once we get it merged I can see what's up with it @bmquinn

config/releases.exs
mbklein: Yes. The notification config is in `sequins.exs` and will be carried over to here. `releases.exs` only needs to override the `queue_config`.

terraform/lambdas.tf
mbklein: Yeah, this should probably be at least 4096 since you're reading the entire thing into memory.

config/releases.exs
bmquinn: Ok thanks, removed.

terraform/lambdas.tf
bmquinn: Bumped to 4096.

mbklein (approved):

lib/meadow/pipeline/actions/extract_exif_metadata.ex
```diff
@@ -0,0 +1,57 @@
...
+  def extract_exif_metadata(url) do
+    case URI.parse(url) do
+      %{host: bucket, path: "/" <> key} -> invoke_lambda(bucket, key)
+      _ -> Logger.error("Cannot extract EXIF from invalid url: #{url}")
+    end
+  end
+
+  defp invoke_lambda(bucket, key) do
+    case Lambda.invoke(Config.lambda_config(:exif), %{bucket: bucket, key: key}) do
+      {:ok, result} -> result
```
kdid, line 53:
```
iex(13)> Lambda.invoke(Config.lambda_config(:exif), %{bucket: bucket, key: key})
[info]  Extracting EXIF metadata: s3://dev-preservation//6b/13/6e/fd/a5acb8dc829138cc0032d0e38a57cebaabe2bf5d90239127eae3706e96b7dfb9
[warn]  Received undefined response from lambda
{:ok, nil}
```


So, our seed data (jpgs) seem to return `undefined` from `exifr`, which `lambda.ex` logs and turns into `{:ok, nil}`. It seems like we're catching the nil in that pattern match as a valid result and going ahead with an attempted db update. Can we change that so it recognizes the undefined/nil?

I'm a little confused why `exifr` returns undefined. With `exiftool` we seem to at least get image width, height, mime type, etc...

terraform/lambdas.tf
```diff
@@ -120,15 +120,23 @@ module "pyramid_tiff_function" {
...
+ memory_size   = 4096
+ timeout       = 600
+
+ environment = {
+    TMPDIR   = "/mnt/working"
+  }
+
+  access_point_arn   = aws_efs_access_point.meadow_working_access_point.arn
+  subnet_ids         = data.aws_subnet_ids.private_subnets.ids
+  security_group_ids = [aws_security_group.meadow_efs_client.id]
```
bmquinn, line 139: @mbklein I'm not sure this is totally correct, but I just used what was in the pyramid lambda. `mnt/working` seems like a catch-all directory but do you think we should split exif and pyramids into their own working dirs?

lib/meadow/pipeline/actions/extract_exif_metadata.ex
bmquinn: @kdid Just pushed up a commit so that case is handled explicitly. I totally agree about the seed data issue, but I think what is happening is that `exifr` is constrained to just pulling the `EXIF` and `JFIF` metadata from the files, whereas `exiftool` does some extra work to pull things like height and width even if they are not explicitly set in the metadata. In the interest of moving towards go-live I think we should probably just move ahead with what we've got and circle back to this once we have some more time.

terraform/lambdas.tf
mbklein: Nah. `tempy` will prevent collisions. And we might eventually have a “download to working dir” action at the beginning of the pipeline and a cleanup action at the end so the digester, exif, and tiff actions can all use that (while still retaining the ability to retrieve it on demand if its missing).

lib/meadow/pipeline/actions/extract_exif_metadata.ex
kdid: Yeah, agree, I didn't mean that you should fix that just the handling of the `nil` response...

kdid (approved):

PR #1707, opened 2021-01-20
Deploy NodeJS lambdas via Terraform

mbklein:
* Create reusable terraform module/manifest for NodeJS lambda deployment
* Deploy `meadow-digester` lambda via terraform
* Configure meadow to use `meadow-digester` lambda in prod with 50 concurrent processors


bmquinn (approved):
Great! Really interesting bit for me is the archive triggering on changes to the zip's `output_sha` If I forget to bring it up, remind me to pick your brain about it later on! 👍

PR #1710, opened 2021-01-21
Add search to NUL Authority dashboard

adamjarling:
Adds search to NUL Dashboard, and created a generic search component wrapper which is being re-used on the Project search as well...and any we want to implement in the future.

image

image


divyareddynu:
This looks great, can we move it to the left? I feel like center is not fitting well within the card.

adamjarling:
@divyareddynu I moved the search bar to the left, and put a `isCentered` optional prop in case we want to center it (since it's re-usable and could be implemented in a variety of ways)

divyareddynu (approved):

PR #1711, opened 2021-01-21
Update Collection page to show facet

divyareddynu:
Screen Shot 2021-01-21 at 9 52 12 AM


adamjarling (changes requested):
Functionality works great, but a few tiny design requests...

image

Also, if my generic Search component gets merged in, could you add that component to the Collections list page as well (since we're in Collections...it'll save creating an extra issue or forgetting about it:) ?

adamjarling (approved):
Looks great!

PR #1722, opened 2021-01-25
Add CSV import dashboard.  Add CSV export from Search screen

adamjarling:
This a partial PR, as it doesn't include tests... I'm back-filling those now, but figured would get this in first.  Everything should be functional and works reasonably well.  

image

image

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ You can pass any public collection ID or valid Elasticsearch query to the downlo
 
 Start the Phoenix with `mix phx.server` or `iex -S mix phx.server` if you want to an interactive shell.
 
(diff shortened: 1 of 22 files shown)


kdid:
Works great. Two things:

- Change "CSV" to "CSV Metadata Update" in the dropdown and breadcrumb since that's what we decided to call this type of operation
- Remove source property display. That is totally behind the scenes to the user and that bucket is not something they have access to. 


Screen Shot 2021-01-25 at 3 04 36 PM

adamjarling:
@kdid Just re-pushed.

kdid (approved):

PR #1726, opened 2021-01-26
Authorized users can create shared links

kdid:


divyareddynu (approved):
Works as expected for USERS

PR #1735, opened 2021-01-27
Added tooltip component and it's test

divyareddynu:
I added a simple tooltip component. This needs to be approved before adding to nulib/

Hidden tooltip:
Screen Shot 2021-01-27 at 8 49 39 AM

Tooltip on hover with content:
Screen Shot 2021-01-27 at 8 49 33 AM


adamjarling:
@divyareddynu On initial test in the UI (w/o looking at the code), hovering makes other screen content jump around.  Could you make the tooltip `position: absolute` relative to the item hovered, to avoid the jumping?

divyareddynu:
@adamjarling That's exactly how it works. Let me see if adding a transition solves this.

adamjarling:
Hmmm, well something is pushing the content on the page around when hovering.  See if you can work out a nice, smooth UX.

divyareddynu:
@adamjarling here's where I tried the styled components from https://emotion.sh/docs/styled

adamjarling (approved):
Looks good.

PR #1736, opened 2021-01-27
Build lambda deploy package during plan stage instead of apply

mbklein:
## Description

Lambda deploy zip files are now Terraform [data sources](https://www.terraform.io/docs/language/data-sources/index.html) instead of [resources](https://www.terraform.io/docs/language/resources/index.html), meaning they will be built on demand and not saved as part of the state. This will prevent missing zip files from causing errors during apply, and also ensure that changes to JS code are picked up more reliably.

## Testing

From the `meadow/terraform` directory
```
$ AWS_PROFILE=staging-admin
$ aws-adfs login --profile=$AWS_PROFILE
$ terraform workspace select staging
$ terraform plan -var-file=staging.tfvars
```
Execution may take a minute while the zips are built, but should come back with
```
No changes. Infrastructure is up-to-date.
```

Then make a small change to one of the lambdas. Maybe just insert an extra carriage return into an `index.js` file or something.

```
$ terraform plan -var-file=staging.tfvars
```

This time it should indicate a change to one `aws_lambda_function` resource, replacing the old filename with the new one.

If you undo the change to the `index.js` file and run `plan` again, it should go straight back to `No changes.`


bmquinn (approved):
All three steps worked as described. Nice!

PR #1741, opened 2021-01-28
Batch Edit View works option

divyareddynu:
Screen Shot 2021-01-28 at 3 41 26 PM
Screen Shot 2021-01-28 at 3 41 35 PM


divyareddynu:
@adamjarling let me know if you like to display `View all works` on the dashboard detail on the screen box instead. Right now, the screen component does not have query and it is rendered from root. I can modify the component to read from params and refetch data. But I think this is a good way to have it under the card details just like how we display on Collections/Ingest Sheet pages.

kdid (changes requested):
Hi @divyareddynu - it looks like you have passed the original query to the search which is not what we want to do here. 

The reason is that that will not necessarily return the works that were affected (metadata that the original query was based on could change in the future). 

What we did was index the batch id on each work in the process of doing the batch edit. So the choices are either:

1) make batch id a facet (but hide the display of the facet from the UI on the search page) and link into from the view all works button.
OR
2) pass the batch id into search to search the batch field

What we were saying in the issue is that we think option number 2 is the better one (simpler and you don't have to deal with hiding the facet and all that stuff)

Let me know if you have any questions....

divyareddynu:
@kdid I'm passing the query from `GET_BATCH`, where can I get the query that updated works otherwise? Do you mean to say search should support searching by batchId?

kdid:
Yep. Search by batch id. Batch id's are indexed on works. You can experiment by manually putting a batch id in the search box. It appears to be configured but you should definitely double check that. 

Screen Shot 2021-01-28 at 4 24 00 PM

adamjarling:
Wow, ok so any time a batch job is executed it applies the `batch job id` to the Work and pushes in onto the existing array of batch ids?   Slick.  Knowing that, yea @divyareddynu we can pass the `batch job id` as a React Router state variable to the Search screen.  The Search screen, and the search box component wrapper itself, should (I think) be listening for any search terms passed in.  If you want to walk through it, lemme know.

adamjarling:
@divyareddynu For the UI, I'd say we pull the link off the "Works Updated" number in the list, and just link to the works from the Details screen.  

For that button, can we label it "View batch edit works"?

divyareddynu:
The tests keep failing on this one probably coz the environment went stale, added a new PR with same edits. Closing this one.

PR #1746, opened 2021-02-01
Create action and lambda to extract mime-type

kdid:
Terraform changes applied on staging - 

example test event:
```
{
  "key": "seedsjan2021-1611852235/BFMF_B34_F16_005_p002.tif",
  "bucket": "meadow-s-ingest"
}
```

expected result:
```
{
  "ext": "tif",
  "mime": "image/tiff"
}
```


lib/meadow/pipeline/actions/extract_mime_type.ex
```diff
@@ -0,0 +1,64 @@
...
+    %{host: bucket, path: "/" <> key} = URI.parse(source)
+
+    case Lambda.invoke(
+           Config.lambda_config(:mime_type),
+           %{bucket: bucket, key: key}
+         ) do
+      {:ok, %{"ext" => _, "mime" => mime_type}} ->
+        {:ok, mime_type}
+
+      {:error, error} ->
```
mbklein, line 54: I'm getting a warning on this line:
```
warning: variable "error" is unused (if the variable is not meant to be used, prefix it with an underscore)
```

mbklein (changes requested):

PR #1752, opened 2021-02-02
Add initial layout to the homepage

adamjarling:
Added elements for general page layout to the Meadow homepage.

**_Note this is draft 1.  See below for latest draft._**

image

### DIFF
--- a/assets/js/components/Collection/RecentlyUpdated.jsx
+++ b/assets/js/components/Collection/RecentlyUpdated.jsx
@@ -1,0 +1,10 @@
+import React from "react";
+import PropTypes from "prop-types";
+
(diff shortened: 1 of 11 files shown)


kdid:
Widgets look good. One thing I noticed - I think we talked about what would happen when logged in as a "User" role and you don't see the "Ingest Objects" action box would the other boxes rearrange, but when I changed my role to user I still see the Ingest box.

kdid:
Looks good but the "+ Add Work" button does not work.

adamjarling:
@kdid Ooops.  Just pushed up the fix.

kdid (approved):

PR #1765, opened 2021-02-03
Fileset form error message bug

divyareddynu:
Screen Shot 2021-02-03 at 9 04 36 AM


kdid (changes requested):
Can we get rid of "FileSet label" and have just "Label" in the error text? (same for description)

adamjarling:
Since we're in this component, can we update the Error message to lose "FileSet" from the Error messages?   Just make "Description..." and "Label..."

image

adamjarling (changes requested):
Since we're in this component, can we update the Error message to lose "FileSet" from the Error messages?   Just make "Description..." and "Label..."

image

adamjarling (approved):
Looks great.

PR #1766, opened 2021-02-03
Add error path to pipeline

kdid:
If an action in the pipeline returns and error, sets the remaining Progress entries for the row to "error", and the remaining ActionState entires for the file set to "error", storing an outcome.

The user will see something like this

Screen Shot 2021-02-03 at 5 20 27 AM


lib/meadow/pipeline/actions/common.ex
```diff
@@ -20,16 +20,23 @@ defmodule Meadow.Pipeline.Actions.Common do
...
+      end
+
+      def update_progress(data, attrs, {status, _, _}, _),
+        do: update_progress(data, attrs, status)
+
+      def update_progress(data, attrs, {status, _}, _), do: update_progress(data, attrs, status)
+      def update_progress(data, attrs, {status, _}), do: update_progress(data, attrs, status)
 
-      def update_progress(%{ingest_sheet: sheet_id, ingest_sheet_row: row_num}, status) do
+      def update_progress(data, %{ingest_sheet: sheet_id, ingest_sheet_row: row_num}, status) do
```
mbklein, line 34: I'm not getting a compiler warning about it (probably because the compiler isn't great at catching things in macros), it looks like `data` can be underscored out here.

lib/meadow/ingest/progress.ex
```diff
@@ -31,6 +31,25 @@ defmodule Meadow.Ingest.Progress do
...
+      on: r.id == p.row_id,
+      where: r.row == ^row_id and p.status == "pending" and r.sheet_id == ^ingest_sheet_id
+    )
+    |> Repo.update_all(set: [status: "error"])
+  end
+
+  @doc """
+  UI upload or case w/out ingest sheet
+  """
+  def abort_remaining_pending_entries(_), do: :noop
```
mbklein, line 51: The compiler is complaining about duplicate `@doc`s – only one `@doc` per combination of function name + arity is allowed. Maybe convert the second one to a comment?

mbklein (approved):

bmquinn (approved):

PR #1769, opened 2021-02-03
Wire up data to homepage components

adamjarling:
The Stats data for the homepage is being calculated almost entirely through the new `useRepositoryStats` hook.  For the Google Analytics component, I updated the theme to pull NU brand colors and NU's fallback font.

image

image


divyareddynu:
Can we add total projects and total ingestSheets?

adamjarling:
@divyareddynu Yea, I'll add that in now.

adamjarling:
@divyareddynu Just updated to include Projects count.  Note it will only count projects which contain a Work.  And I don't think there's a way to pull Inventory Sheet count from the Elasticsearch index.  I'd say don't worry about wiring that up in your issue for now.  We'll add it back in later if we really want/need it.

kdid:
Looks good. I guess we should display the labels for visibility: "Public", "Institution" and "Private", (rather than "Open", "Authenticated" and "Restricted"). That way they will match up with the tags on the site too.

adamjarling:
@kdid Good call.  Just pushed an update.

kdid (approved):

PR #1774, opened 2021-02-04
Allow logged in admins to change their role temporarily

kdid:
- Let's an administrator drop down their role temporarily from the navbar until they logout.

<hr>

Screen Shot 2021-02-04 at 9 41 45 AM

Screen Shot 2021-02-04 at 9 41 37 AM

### DIFF
--- a/assets/js/components/Role/NavDropdown.jsx
+++ b/assets/js/components/Role/NavDropdown.jsx
@@ -1,0 +1,80 @@
+import React, { useState, useContext } from "react";
+import { useMutation } from "@apollo/client";
+import { AuthContext } from "@js/components/Auth/Auth";
+import UILayoutNavDropdown from "@js/components/UI/Layout/NavDropdown";
+import UILayoutNavDropdownHeader from "@js/components/UI/Layout//NavDropdownHeader";
+import UILayoutNavDropdownBody from "@js/components/UI/Layout/NavDropdownBody";
+import UILayoutNavDropdownItem from "@js/components/UI/Layout/NavDropdownItem";
+import { ASSUME_ROLE } from "@js/components/Role/role.gql.js";
+import { GET_CURRENT_USER_QUERY } from "@js/components/Auth/auth.gql";
+import { toastWrapper } from "@js/services/helpers";
+
+const RoleNavDropdown = () => {
+  const currentUser = useContext(AuthContext);
+  const [activeHoverNav, setActiveHoverNav] = useState("adam");
+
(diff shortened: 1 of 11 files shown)


bmquinn (approved):
Works great! 👍

mbklein (approved):

assets/js/components/Role/NavDropdown.jsx
```diff
@@ -0,0 +1,80 @@
...
+import UILayoutNavDropdownHeader from "@js/components/UI/Layout//NavDropdownHeader";
+import UILayoutNavDropdownBody from "@js/components/UI/Layout/NavDropdownBody";
+import UILayoutNavDropdownItem from "@js/components/UI/Layout/NavDropdownItem";
+import { ASSUME_ROLE } from "@js/components/Role/role.gql.js";
+import { GET_CURRENT_USER_QUERY } from "@js/components/Auth/auth.gql";
+import { toastWrapper } from "@js/services/helpers";
+
+const RoleNavDropdown = () => {
+  const currentUser = useContext(AuthContext);
+  const [activeHoverNav, setActiveHoverNav] = useState("adam");
```
adamjarling, line 14: Did I accidentally leave this in previously?  Hopefully code is not cueing off of  a state value of `adam`?

adamjarling (commented):
@kdid Just a thought, instead of it's own primary menu item, could this visually live under the User top level nav item?  It doesn't feel like a primary action to me, but rather related to User administration in some fashion.   I'd be happy to make the UI updates if you want too.

image

assets/js/components/Role/NavDropdown.jsx
kdid: lol. yeah I copied and pasted it from here. https://github.com/nulib/meadow/blob/0fd27eb547d897d59e9dd4294bd53f2843f19ec1/assets/js/components/UI/Layout/NavBar.jsx#L24 I wasn't sure of the secret meaning. 

But yeah, I think I could make it a submenu of user. I think that' what you're saying. I'll look at the Bulma docs a bit.

kdid: I'll remove both "adam"s

adamjarling: And, not to create more work, but looks like Bulma's CSS and markup for displaying a nav dropdown item and list might have changed slightly from when we implemented it?  (In case things don't seem to display quite right)

kdid:
Updated:

Screen Shot 2021-02-04 at 1 00 43 PM

adamjarling (approved):

PR #1777, opened 2021-02-04
GenerateFileSetDigests pattern matches against binary-size

bmquinn:
- Ensures the `sha256` and `sha1` hashes are the correct binary-size when passed into `FileSets.update_file_set/2`

### DIFF
--- a/lib/meadow/pipeline/actions/generate_file_set_digests.ex
+++ b/lib/meadow/pipeline/actions/generate_file_set_digests.ex
@@ -25,15 +25,21 @@ defmodule Meadow.Pipeline.Actions.GenerateFileSetDigests do
     try do
       Logger.info("Generating digests for #{file_set.id}")
 
-      hashes = generate_hashes(file_set.metadata.location)
-      Logger.info("Hash for #{file_set.id}: #{hashes.sha256}")
+      case generate_hashes(file_set.metadata.location) do
+        {:ok,
+         %{"sha256" => <<sha256::binary-size(64)>>, "sha1" => <<_sha1::binary-size(40)>>} = hashes} ->
+          Logger.info("Hash for #{file_set.id}: #{sha256}")
+          {:ok, _} = FileSets.update_file_set(file_set, %{metadata: %{digests: hashes}})
+          ActionStates.set_state!(file_set, __MODULE__, "ok")
+          :ok
 
-      {:ok, _} =
-        file_set
-        |> FileSets.update_file_set(%{metadata: %{digests: hashes}})
+        {:ok, unexpected_result} ->
+          ActionStates.set_state!(file_set, __MODULE__, "error", unexpected_result)
+          {:error, unexpected_result}
 
-      ActionStates.set_state!(file_set, __MODULE__, "ok")
-      :ok
+        {:error, error} ->
+          raise error
+      end
     rescue
       Meadow.Utils.Stream.Timeout ->
         ActionStates.set_state!(file_set, __MODULE__, "pending")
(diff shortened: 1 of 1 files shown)


lib/meadow/pipeline/actions/generate_file_set_digests.ex
```diff
@@ -25,15 +25,21 @@ defmodule Meadow.Pipeline.Actions.GenerateFileSetDigests do
...
+      case generate_hashes(file_set.metadata.location) do
+        {:ok,
+         %{"sha256" => <<sha256::binary-size(64)>>, "sha1" => <<_sha1::binary-size(40)>>} = hashes} ->
+          Logger.info("Hash for #{file_set.id}: #{sha256}")
+          {:ok, _} = FileSets.update_file_set(file_set, %{metadata: %{digests: hashes}})
+          ActionStates.set_state!(file_set, __MODULE__, "ok")
+          :ok
+
+        {:ok, unexpected_result} ->
+          ActionStates.set_state!(file_set, __MODULE__, "error", unexpected_result)
```
kdid, line 37: Does the `unexpected_result` need to be converted to a string or something in order to be stored? I don't think we know what it is?

lib/meadow/pipeline/actions/generate_file_set_digests.ex
```diff
@@ -25,15 +25,21 @@ defmodule Meadow.Pipeline.Actions.GenerateFileSetDigests do
...
+          {:error, unexpected_result}
+
+        {:error, error} ->
```
kdid, line 40: I feel like we need to handle the case where the return is `_` (anything else) and store an error for this action

lib/meadow/pipeline/actions/generate_file_set_digests.ex
bmquinn: It ends up looking like:
```
action_states: [
    %Meadow.Data.Schemas.ActionState{
      __meta__: #Ecto.Schema.Metadata<:loaded, "action_states">,
      action: "Meadow.Pipeline.Actions.GenerateFileSetDigests",
      id: "9d204ba7-d897-416a-b739-6b9ed8f9d76e",
      inserted_at: ~U[2021-02-04 20:10:57.965955Z],
      notes: "%{sha1: nil, sha256: nil}",
      object_id: "650b2a56-5aac-4fc7-891f-a306081a9af1",
      object_type: "Meadow.Data.Schemas.FileSet",
      outcome: "error",
      updated_at: ~U[2021-02-04 20:10:58.543300Z]
    }
  ]
  ```

File: lib/meadow/pipeline/actions/generate_file_set_digests.ex (hunk 2 above)
Comment (bmquinn) [reply lib/meadow/pipeline/actions/generate_file_set_digests.ex:40]: I think that's handled between the `{:ok, unexpected_result}` case and the `{:error, error}` case, right?

File: lib/meadow/pipeline/actions/generate_file_set_digests.ex (hunk 1 above)
Comment (kdid) [reply lib/meadow/pipeline/actions/generate_file_set_digests.ex:37]: Yeah, I think we're fine on this one since MBK's PR got merged.

File: lib/meadow/pipeline/actions/generate_file_set_digests.ex (hunk 2 above)
Comment (kdid) [reply lib/meadow/pipeline/actions/generate_file_set_digests.ex:40]: We might. It looks like `Lambda.invoke` could return `other -> other` and I don't think we have any idea what that is. It might not be covered with our cases, right?

Comment (mbklein) [reply lib/meadow/pipeline/actions/generate_file_set_digests.ex:40]: I guess there is a tiny chance that we could get back something that doesn't match either `{:ok, ...}` or `{:error, ...}`. Maybe a case for

```
very_unexpected_result -> 
    ActionStates.set_state!(file_set, __MODULE__, "error", very_unexpected_result)
    {:error, very_unexpected_result}
```

??

mbklein: But wait... `Meadow.Utils.Lambda` is already making sure that everything is coming back as `{:ok, something}` or `{:error, something}` by contract. We don't need the third case.

kdid: where exactly is this happening @mbklein ?

PR #1778, opened 2021-02-04
Catch the ArgumentError raised by Pairtree.preservation_path()

kdid:


mbklein (commented):
Can there be a test case for this? Just a copy of the same `success` case currently defined in `copy_file_to_preservation_test.exs` but with a invalid sha256 in the file set fixture.

PR #1798, opened 2021-02-09
Eliminate temp storage in exif and pyramid TIFF lambdas

mbklein:
I made some changes to the Pyramid TIFF and EXIF lambdas to eliminate the need for temporary storage.

### Changes
* The pyramid lambda now pipes the input stream from S3 directly to sharp, which transforms the stream in memory and pipes it into a stream that uploads it to the destination
* The EXIF lambda now reads the entire file into a memory buffer and extracts the EXIF from there

### Pros
No more temp storage

### Cons
Both actions are now memory-bound, but:
* Lambdas are now configurable up to 10GB (up from 3GB when it was first determined that the TIFF processor needed temp storage)
* Both lambdas should already be configured with plenty of memory – 8G for TIFF; 4G for EXIF – to handle the largest source files we've been talking about

### DIFF
--- a/priv/nodejs/pyramid-tiff/pyramid.js
+++ b/priv/nodejs/pyramid-tiff/pyramid.js
@@ -1,81 +1,68 @@
 const AWS = require("aws-sdk");
+const concat = require("concat-stream");
 const sharp = require("sharp");
 const URI = require("uri-js");
-const fs = require("fs");
-const tempy = require("tempy");
+const s3 = new AWS.S3();
 
-const createPyramidTiff = async (source, dest) => {
-  const inputFile = await makeInputFile(source);
-  try {
-    console.info(`Creating pyramidal TIFF from ${inputFile}`);
-    const pyramidTiff = await sharp(inputFile, { limitInputPixels: false })
-      .resize({
-        width: 15000,
-        height: 15000,
-        fit: "inside",
-        withoutEnlargement: true
-      })
-      .tiff({
-        compression: "jpeg",
-        quality: 75,
-        tile: true,
-        tileHeight: 256,
-        tileWidth: 256,
-        pyramid: true
-      })
-      .toBuffer();
-    await sendToDestination(pyramidTiff, dest);
-    return dest;
-  } finally {
-    console.info(`Deleting ${inputFile}`);
-    fs.unlink(inputFile, err => {
-      if (err) {
-        throw err;
-      }
-    });
-  }
-};
+AWS.config.update({httpOptions: {timeout: 600000}});
 
-const makeInputFile = location => {
+const createPyramidTiff = (source, dest) => {
   return new Promise((resolve, reject) => {
-    let uri = URI.parse(location);
-    let fileName = tempy.file();
-    console.info(`Retrieving ${location} to ${fileName}`);
-    let writable = fs
-      .createWriteStream(fileName)
-      .on("error", err => reject(err));
-    let s3Stream = new AWS.S3()
-      .getObject({
-        Bucket: uri.host,
-        Key: getS3Key(uri)
-      }).createReadStream();
+    try {
+      console.log(`Creating pyramid from ${source}`);
+      const inputStream = streamFromS3(source);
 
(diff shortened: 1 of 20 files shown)


bmquinn:
I'm 💯 for this change. Removing the EFS mount simplifies these lambdas tremendously, and since we can work within the (now higher) memory limits this looks like the best way forward. We can deal with use cases that fall out of this range if/when they come up down the line, possibly with a different approach such as S3 event notifications. Since this change can be terraform-applied we should also test these lambdas tomorrow on staging before un-drafting or merging this PR.

priv/nodejs/pyramid-tiff/pyramid.js
```diff
@@ -1,81 +1,65 @@
...
-      .getObject({
-        Bucket: uri.host,
-        Key: getS3Key(uri)
-      }).createReadStream();
+    try {
+      console.log(`Creating pyramid from ${source}`);
+      const inputStream = streamFromS3(source);
+
+      let metadata;
+      const transformStream = sharp({ limitInputPixels: false })
```
bmquinn, line 14: Should we test chaining the `removeAlpha` operation here as well? We've been bitten by alpha channels many times, and I just came across it in the [docs](https://sharp.pixelplumbing.com/api-channel#removealpha).

kdid:
Agree we should do this and avoid EFS altogether.

I'm wondering, did we intentionally lose the `AWS.config.update({httpOptions: {timeout: 600000}});`  ? (applied but not PR'd)
It did help with the timeouts. Maybe you're thinking the new code doesn't need it but the digester didn't change, and does it hurt anything?



Note: I do see a *ton* of this in the logs:

```
action=CreatePyramidTiff [warn]  Skipping Elixir.Meadow.Pipeline.Actions.CreatePyramidTiff for 4dff6680-9e94-4929-a09a-66a9685f68c0 – already complete
action=ExtractExifMetadata [warn]  Skipping Elixir.Meadow.Pipeline.Actions.ExtractExifMetadata for 65436ffe-4a46-4b9b-9422-ba53df59ccb9 – already complete
```

and I so also see some ExAws timeouts. (I switched some time ago to seeding with tifs (not huge ones) - since jpgs don't work the same in the pipeline)

```
action=CopyFileToPreservation [warn]  ExAws: HTTP ERROR: :timeout for URL: "https://devbox.library.northwestern.edu:9001/dev-preservation/af/cb/7c/fc/afcb7cfc2317d86cab897a0a1fdcdc89717bd4ba626d74fd47b68fb50e15a28e" ATTEMPT: 4
```

mbklein:
Reverting the AWS timeout in the lambda was not intentional; I just used `deploy/staging` as the source branch and forgot. I have one other sharp option I forgot to add as well, so I’ll push an update before removing from draft status.

mbklein:
I did some testing with a new lambda, first with the EXIF code and then with the pyramid code, to see how much memory/time it took to complete successfully with the [2.5GB Mars Rover TIFF](https://mars.nasa.gov/resources/curiositys-1-8-billion-pixel-panorama/?site=msl) I’ve been using as “the big one.” 

Neither one completed successfully with 3072 MB or 4096 MB. (We’re not tied to GB boundaries, but they make for easy and obvious choices to test with.)

With 8192 MB:
* EXIF completed in 33898ms (~34 seconds) with a peak memory usage of 4803 MB
* TIFF completed in 115922ms (~2 minutes) with a peak memory usage of 5241 MB

I think it'd be safe to set both of these to 6144 MB with 240 second (4 minute) timeouts and then tweak as necessary.

We can also – NOT THIS WEEK – contemplate the idea of deploying multiple copies with different RAM configurations, and choosing which one to call at runtime based on the size of the data.

davidschober:
Thanks Michael. I also found 
https://github.com/alexcasalboni/aws-lambda-power-tuning recommended by the folks at amazon to gather data in helping to find sweet spots. It seems that with the exif lambda, we will gain a lot from streaming as I can't image the operation is ram or CPU intensive. I'll kick off an issue on the board to implement whenever it's sorted upstream.

PR #1803, opened 2021-02-09
Beginning of a strategy on how to determine the contents of a manifes…

adamjarling:
…t are different.  Not happy with the approach, but here it should load fresh content on a page refresh

See issue comments for explanation of where this thing is at.

https://github.com/nulib/repodev_planning_and_docs/issues/1346

### DIFF
--- a/assets/js/components/Work/Work.jsx
+++ b/assets/js/components/Work/Work.jsx
@@ -12,15 +12,48 @@ const osdOptions = {
 };
 
 const Work = ({ work }) => {
+  const [manifestObj, setManifestObj] = React.useState();
+  const [randomUrlKey, setRandomUrlKey] = React.useState(Date.now());
+  const [manifestChangeItems, setManifestChangeItems] = React.useState({
+    label: "",
+    canvasCount: "",
+  });
+
+  React.useEffect(() => {
+    async function getData() {
+      const response = await fetch(`${work.manifestUrl}?${Date.now()}`);
+      const data = await response.json();
+
+      // Check if watch items in manifest are different.
+      // If so, trigger a re-render of OSD viewer
+      if (
+        data.label !== manifestChangeItems.label ||
+        data.sequences[0].canvases.length !== manifestChangeItems.canvasCount
+      ) {
+        setManifestChangeItems({
+          label: data.label,
+          canvasCount: data.sequences[0].canvases.length,
+        });
+        setRandomUrlKey(Date.now());
+      }
+      setManifestObj(data);
+    }
+
+    getData();
+  }, [work.fileSets, work.descriptiveMetadata.title]);
+
   return (
     <>
       <section>
         <div data-testid="viewer">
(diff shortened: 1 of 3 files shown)


bmquinn:
@adamjarling I had an idea that hopefully might help. Since the manifest is an S3 object, the `etag` will change every time the manifest is updated so you would store the `etag` in the component state and check against the `etag` stored in the state using the `If-None-Match` header. A match (i.e. an unchanged manifest) returns a `304 Not Modified` and a mismatch (updated manifest) would return a `200 OK` with the new document. This way you wouldn't have to know anything about a Work or its FileSets changing within the manifest itself. Not sure if React's `fetch` API provides this kind of functionality or if you'd have to write this logic in the component, but I thought I'd throw it out there.

Here's an example using curl:

```sh
etag=1a25ac1db61c4d626fe50278400b8369
url=https://stack-s-pyramid-tiffs.s3.amazonaws.com/public/2a/02/62/2c/-3/e4/3-/4f/5c/-8/26/f-/de/ee/17/eb/19/DELETE-0b-manifest.json

curl -I $url --header 'If-None-Match: '"$etag"''
HTTP/1.1 304 Not Modified
Last-Modified: Tue, 09 Feb 2021 23:26:30 GMT
ETag: "1a25ac1db61c4d626fe50278400b8369"

curl -I $url --header 'If-None-Match: '"$etag"''
HTTP/1.1 200 OK
Date: Tue, 09 Feb 2021 23:56:12 GMT
Last-Modified: Tue, 09 Feb 2021 23:56:05 GMT
ETag: "6d60c31413db0f2a8c16af8b89c65232"
Accept-Ranges: bytes
Content-Type: application/json
Content-Length: 1530
```

adamjarling:
@bmquinn Sweet, thanks for the tip.  Something like the `etag` is exactly what the component would need to evaluate if the manifest is different (instead of my current approach of actually pulling down the info and manually checking). 

Unfortunately I think timing is still going to be the trick here.  A component re-renders when some data changes that's passed into the component (ie. GraphQL query data, or a parent-level prop value).   This data which triggers a component re-render happens before the new manifest contents are updated (so it seems).  So we'd either need polling (which I messed with via `setInterval` or some type of message to grab onto that the file has changed.   I was polling it 5 times at 1000ms intervals, and something about that just felt wrong.  

I'll come back to this if it appears to be a deal breaker for a proper solution.

bmquinn:
@adamjarling Looks like you've got merge conflicts checked into `yarn.lock`. Probably need to clobber it and regenerate.

adamjarling:
@bmquinn Yea I think I fixed these yesterday before seeing your message, but just re-based on `deploy/staging` and re-pushed.

kdid:
@adamjarling  Is this ready for review? I got confused on the status.

adamjarling:
@kdid Yes, it could get merged and then if we want to build off it later, we can.

kdid (approved):

PR #1817, opened 2021-02-11
Change batch query_string to search non-tokenized id field

kdid:
The batch edit/delete query (in select works mode), searches on the `id` field, which is tokenized. So a search for an id of 
bdfa8b37-5f43-4594-a8f4-4bd26a27900e

would also match this id because it contains a8f4
7916a549-b921-4318-a8f4-3f2386732bb5

This would lead someone to edit or delete more works than they thought they were.


adamjarling (approved):
Good catch.

PR #1836, opened 2021-02-16
Change ReactiveSearch search sensor from SearchSensor to q

adamjarling:
Could someone verify I should update this Elixir Test file?   Just did a search and replace and came across it.

image


kdid:
Yep. And it looks like the (Elixir) tests are passing with the change.

kdid (approved):

PR #1838, opened 2021-02-16
Update Ingest sheet page details info and general cleanup

adamjarling:
@vlsrobinson we can always update this more in the future, but not sure how valuable some of the info would be?

image

### DIFF
--- a/assets/js/components/IngestSheet/Completed.jsx
+++ b/assets/js/components/IngestSheet/Completed.jsx
@@ -11,10 +11,12 @@ import IngestSheetCompletedErrors from "./Completed/Errors";
 import UIPreviewItems from "../UI/PreviewItems";
 import UISkeleton from "../UI/Skeleton";
 import { Button } from "@nulib/admin-react-components";
-import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import IngestSheetDetails from "@js/components/IngestSheet/Details";
+import { elasticsearchDirectSearch } from "@js/services/elasticsearch";
 
 const IngestSheetCompleted = ({ sheetId, title }) => {
   const history = useHistory();
+  const [totalWorks, setTotalWorks] = React.useState();
   const {
     loading: worksLoading,
     error: worksError,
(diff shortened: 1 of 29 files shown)


kdid:
I ingested 500 works but it says Total Works 10, which is the number it's previewing.

adamjarling:
Ah, one sec.

vlsrobinson:
Looks great! I think the 'complete', preview of works and total works created is  good for now.  We can see if the Prod team wants additional info at some point in the future after we get it in front of them (like # of file sets or something.)  But I would just want it as a quick double-check.

adamjarling:
@kdid So the 10 value is coming back because we're limiting the results as part of the GraphQL query.  https://github.com/nulib/meadow/blob/9a2496a19dab762667c8af88922957c6cca3b281/assets/js/components/IngestSheet/Completed.jsx#L23

And we put this in place because returning all the rows was breaking the UI in big number scenarios, right?  So we need some way to get an aggregate count of this from GraphQL, without returning the data.   Looks like GraphQL doesn't support this out of the box, but maybe we create a new Query which just lists the total count?  

I'm not seeing any obvious place we can get this data at the moment?   Maybe we just suppress the "Details" section for now?

adamjarling:
Or maybe we grab it from ElasticSearch for successful ingests?

kdid:
Yes, can get that from Elasticsearch

assets/js/components/IngestSheet/Completed.jsx
```diff
@@ -29,6 +31,37 @@ const IngestSheetCompleted = ({ sheetId, title }) => {
...
+            ],
+          },
+        },
+        size: 0,
+      });
+      setTotalWorks(response.hits.total);
+    };
+    //fn();
+  }, [sheetId]);
+
```
adamjarling, line 64: @kdid This is where we're now grabbing the info direct from Elasticsearch.

kdid: Is it working for you locally? I'm seeing your getting results from Elasticsearch but I'm not seeing it show in the UI.

kdid: also `yarn.lock` conflict

adamjarling: @kdid Just fixed up both now.  I think I messed up a local stash before the last push, or something weird.  Everything should display now and blew away the `yarn.lock` for a fresh copy.

kdid (approved):

PR #1842, opened 2021-02-17
Delete a fileset functionality from Work Preservation tab

adamjarling:
Delete functionality now wired up.  Handles errors and updates the Apollo Client cache after a delete.

image


bmquinn:
Just running through this PR and the delete worked great, but I noticed the IIIF manifest isn't getting updated to remove the file_sets. Is that out of scope for this issue?

Would expect the following IIIF manifest to have no file_sets:
deleted_file_sets

kdid (approved):

PR #1855, opened 2021-02-19
Terraform tweaks for production

mbklein:
* Bring the Meadow EC2 instance under terraform control
* Fix a few other settings to make sure everything comes up cleanly from scratch

These changes have already been deployed to production via `terraform apply`


bmquinn (approved):
Wondering about ways to make sure we don't forget to update `terraform/ec2_files/dev.local.exs ` when we make config changes, maybe just a note in the documentation would do it but I can foresee forgetting to update biting us down the line.

PR #1858, opened 2021-02-19
Force HTTPS if the request is coming from a load balancer or proxy

mbklein:


lib/meadow_web/plugs/require_https_proxy.ex
```diff
@@ -0,0 +1,27 @@
...
+  @behaviour Plug
+
+  def init(opts), do: opts
+
+  def call(conn, _default) do
+    if from_insecure_proxy?(conn), do: redirect_to_https(conn), else: conn
+  end
+
+  defp from_insecure_proxy?(conn) do
+    get_req_header(conn, "x-forwarded-proto") |> Enum.member?("http")
```
bmquinn, line 16: According to the [docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html) I think you're going to want to check the request for `X-Forwarded-Proto` rather than the all lowercased string. I switched it to the capitalized version in the test's conn setup to check and they failed, too.

bmquinn (changes requested):

lib/meadow_web/plugs/require_https_proxy.ex
mbklein: AWS does capitalize its headers, but Plug.Conn [converts them all to lowercase](https://hexdocs.pm/plug/Plug.Conn.html) for consistency. 

> Note request headers are normalized to lowercase and response headers are expected to have lowercase keys.

bmquinn: Right on, thanks for the link!

bmquinn (approved):

PR #1875, opened 2021-02-23
Create a more robust mechanism for Indexing and updating IndexTimes

mbklein:
- Update and Insert separately instead of trying to use `insert_all` with `on_conflict: update`
- Add index count logging to Indexer
- Increase chunk size
- Reduce wait interval
- Increase timeouts
- Handle Elasticsearch upload failures (trap, warn, move on)

I ran a full `Indexer.reindex_all()` on staging with these changes and it ran through just fine (78 Collections + 80,541 Works + 81,284 FileSets = 161,903 IndexTime records)


kdid (approved):

bmquinn (approved):
Looks great!

PR #1876, opened 2021-02-23
Configure honeybadger on build

kdid:
- Configures Honeybadger on build
- Configures Honeybadger for react app - in the most basic way, there is a lot more that will be customized. 
-  Puts the user in the context as well.

This will report all errors to Honeybadger. 
To test in your dev environment, grab the API key from nulib/tfvars repo: staging.tfvars file and use configure that and an environment (don't use "dev" or "test") in `app.js`, then cause an error. Then you can see the errors at https://app.honeybadger.io/

Screen Shot 2021-02-23 at 11 47 21 AM
Screen Shot 2021-02-23 at 12 02 26 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,7 +128,7 @@ jobs:
       - docker/check
       - docker/build:
           cache_from: nulib/meadow-deps:${DEPLOY_TAG}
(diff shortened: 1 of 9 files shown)


mbklein (approved):
Looks great! We should probably make the back-end as information rich as the front-end (i.e., add revision and logged-in user info to the Elixir honeybadger client also), but that can be a separate back-end ticket.

PR #1880, opened 2021-02-24
Fix HTTP->HTTPS redirect

mbklein:
Replace network load balancer with application load balancer
Rewrite iex-remote script not to go through load balancer

Terraform only update. Already applied on staging and production.


kdid:
Should Meadow be working? I'm getting redirected but seeing "503 Service Temporarily Unavailable"

PR #1886, opened 2021-02-24
Link the collection name on Admin tab to search results

kdid:


assets/js/components/Work/Tabs/Administrative.jsx
```diff
@@ -190,9 +200,15 @@ const WorkTabsAdministrative = ({ work }) => {
...
-                        {collection.title}
-                      </Link>
+                      <Button
+                        className="button is-text"
+                        onClick={() =>
+                          handleViewAllWorksClick(collection.title)
+                        }
+                        data-testid="view-collection-works-button"
+                      >
+                        <span>{collection.title}</span>
```
adamjarling, line 210: No big deal, but do we need this `span` here?

assets/js/components/Work/Tabs/Administrative.test.jsx
```diff
@@ -69,6 +69,8 @@ describe("Work Administrative tab component", () => {
       expect(getByText(/Another Project Description/i)).toBeInTheDocument();
       expect(getByText(/Project Cycle Name/i)).toBeInTheDocument();
       expect(getByText(/Started/i)).toBeInTheDocument();
+      expect(getByText(/Collection 1232432 Name/i)).toBeInTheDocument();
```
adamjarling, line 72: Just to refine a pattern, we don't need the `.toBeInTheDocument()` chained function anymore.  `expect` will handle the truthiness of the return value.  Mind wiping those off this test?   I'm doing it as I come across tests which still use it.

adamjarling (changes requested):
All looks great.  Just a few minor thoughts.

adamjarling (approved):

PR #1894, opened 2021-02-24
Configure Honeybadger with front end

mbklein:
This commit [was reverted](https://github.com/nulib/meadow/pull/1893) because it was preventing the front-end from rendering without a valid login. (This might not be the error condition, but I didn't have a problem until I started a new session and was no longer logged in.)

Cherry-picking it back in and reopening it so it can be iterated on.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,7 +128,7 @@ jobs:
       - docker/check
       - docker/build:
           cache_from: nulib/meadow-deps:${DEPLOY_TAG}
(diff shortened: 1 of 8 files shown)


adamjarling:
@mbklein  This should be good to go now.

adamjarling:
Looks like for non-logged in users it was not catching errors in the Auth component, and also in the `useIsAuthorized` hook on some function definitions which I had outside the actual hook function (for organization), but tucked those inside the hook function and they're not firing for non-logged in users anymore.

adamjarling (approved):

PR #1899, opened 2021-02-25
Update role-based viewing in Meadow UI

adamjarling:
Role viewing updated per the ticket.  So basically in the Search screen, previously the main action buttons were hidden for EDITOR and below, but now display the buttons and wrap the UI show/hide logic around the individual buttons themselves.

image


kdid:
So, editors can export metadata but just not import, correct?

adamjarling:
@kdid Yea that's the way I left it (figured no harm in exporting), but can disable if @davidschober thinks so?

PR #1903, opened 2021-02-26
Change file set role to coded term

kdid:
- Change file set roles to coded terms rather than strings
- "am" and "pm" changed to "A" and "P" in the ingest sheet's role column. 
- updated pipeline, migration, dc_download, ingest
-  was concerned about file set ordering but it still seems to work

`devstack down -v` required and also your current dc_downloads won't work, you'll need to redo them (or manually change "am" -> "A" and "pm" -> "P" - but it would be great if you help me test this)

### DIFF
--- a/assets/js/components/Work/Tabs/Preservation/FileSetForm.jsx
+++ b/assets/js/components/Work/Tabs/Preservation/FileSetForm.jsx
@@ -3,10 +3,11 @@ import PropTypes from "prop-types";
 import UIFormInput from "@js/components/UI/Form/Input.jsx";
 import UIFormField from "@js/components/UI/Form/Field.jsx";
 import UIFormSelect from "@js/components/UI/Form/Select.jsx";
-import { FILE_SET_ROLES } from "@js/services/global-vars";
-import Error from "@js/components/UI/Error";
+import { useCodeLists } from "@js/context/code-list-context";
 
 function WorkTabsPreservationFileSetForm({ s3UploadLocation }) {
+  const codeLists = useCodeLists();
+
   return (
     <>
       {s3UploadLocation && (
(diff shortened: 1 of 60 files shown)


mbklein (changes requested):
I tested two ingest sheets, manually added a couple file sets, and clicked around the UI to make sure everything looked right, and it did. I only have one migration-related change request.

lib/meadow/migration.ex
```diff
@@ -297,6 +298,10 @@ defmodule Meadow.Migration do
     end
   end
 
+  defp update_role(%{role: _role} = file_set) do
+    put_in(file_set.role, %{id: "A", scheme: "FILE_SET_ROLE"})
```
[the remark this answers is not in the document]
kdid: we decided this is OK

assets/js/client-local.js
```diff
@@ -160,3 +160,11 @@ export const mockWorkType = () => {
     __typename: "CodedTerm",
   };
 };
+
+export const mockFileSetRole = () => {
```
adamjarling, line 164: Are we using this anywhere?

assets/js/components/Work/Tabs/Preservation/FileSetForm.jsx
```diff
@@ -3,56 +3,61 @@ import PropTypes from "prop-types";
...
-            <UIFormInput
-              isReactHookForm
-              required
-              label="Accession number"
-              data-testid="fileset-accession-number-input"
-              name="accessionNumber"
-              placeholder="accession number"
-            />
-          </UIFormField>
+          <CodeListProvider>
```
adamjarling, line 16: Not sure we need the `CodeListProvider`?   Seems to work without it but probably not a big deal if it's not throwing errors.

adamjarling (approved):
Front end looks good to me.  Nice work!

mbklein (approved):

mbklein:
There is a bug in the `CopyFileToPreservation` action that breaks migration on this branch, but it's fixed in #1904.

bmquinn (approved):

PR #1909, opened 2021-02-26
Use advisory locks in ingest sheet progress trigger

mbklein:
This should resolve the race condition that sometimes prevents the ingest sheet from being marked `completed`.


mbklein:
Nope. We can just run a targeted rollback/migrate that only replaces the trigger while everything else remains unchanged.

kdid (approved):

PR #1925, opened 2021-03-02
Move update Collection Work to Work Administrative tab

adamjarling:
This moves the setting a Collection representative Work to the Work screen.

image

Since the API is set up as only a one-way operation (an error throws sending a blank value to the `setCollectionImage` mutation for `workId`), blocking removing a Work as a Collection work like this:

image


kdid:
It works well to set the image, but I feel like there needs to be a way to get to and/or identify the collection's current representative work?

Like maybe a link from here somewhere? :

Screen Shot 2021-03-02 at 1 53 49 PM

adamjarling:
@kdid Not sure I follow, but think I've lost perspective on this thing after fighting with it for 2 days.

kdid:
Like let's say you wanted to look at the work itself that is the representative work. (For example maybe you want to make sure which work it is as the images are sometimes similar/identical. Or let's say you wanted to change the work's representative image.) How do you get to the page that is the current representative work for the collection?

adamjarling:
Ahhh.. yea I get it.  Good point.  I'll add a link from the Collection page.

adamjarling:
@kdid Just added a link on the Collection image itself (if it has a representative image)

kdid (approved):

PR #1938, opened 2021-03-04
Change total works count for ingest sheet to GraphQL query

kdid:


assets/js/components/IngestSheet/Completed.jsx
```diff
@@ -84,7 +61,7 @@ const IngestSheetCompleted = ({ sheetId, title }) => {
   return (
     <div>
       {ingestSheetErrors.length === 0 && (
-        <IngestSheetDetails totalWorks={totalWorks} />
+        <IngestSheetDetails totalWorks={workCountData.ingestSheetWorkCount} />
```
adamjarling, line 64: Looks great.  Would there ever not be a `ingestSheetWorkCount`, or will it always return something (`null` or `0`)?

adamjarling (approved):

assets/js/components/IngestSheet/Completed.jsx
kdid: Query returns an integer, so it would be zero. Is that an issue?

Screen Shot 2021-03-04 at 8 30 17 AM

adamjarling: As long as it always returns some type of truthy value (even though it could be a 0 here), then yea should be all good.

PR #1940, opened 2021-03-04
Add robust semantic versioning support

mbklein:
- Integrate Meadow version with Honeybadger
- Include version in context on all Honeybadger notifications
- Update CircleCI to pass meadow version to Docker
- Update Dockerfile to pass meadow version to webpack
- Update webpack config to include version
- Configure CircleCI to tag master branch with version and push to github
- Create ADR for semantic versioning

While checking to see if webpack handled the `MEADOW_VERSION` properly, I found that it wasn't precompiling any of the `process.env.*` variables, which led me to the [`DefinePlugin`](https://webpack.js.org/plugins/define-plugin/) solution. Apparently `webpack` handles the build-time environment differently from however `react-scripts` does it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,10 +125,14 @@ jobs:
             else
               echo "export DEPLOY_TAG=$(echo $CIRCLE_BRANCH | awk -F/ '{print $NF}')" >> $BASH_ENV
             fi
+      - run:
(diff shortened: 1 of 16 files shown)


adamjarling (approved):
Looks good to me.

PR #1957, opened 2021-03-08
Build source maps for application code

mbklein:
Because webpack's minimizer has been configured with `sourceMap: false`, maps were only being generated for the vendor bundle, not the app bundle. This change tells webpack to build the source map for the app bundle as well.


adamjarling (approved):
Nice!

PR #1958, opened 2021-03-08
Blanket UI cleanup

adamjarling:
For this, just took a time-boxed trip through the app and tried to do a few things:

- Update anything that has been generally sticking out to me
- Try to use some of the NU marketing headline styles
- Upgrade Icons where applicable

I didn't touch any notifications or tags yet since there are separate issues for that.


kdid:
Looks good, the only thing I noticed is that the link to the collection on top of the work page should go to search result (like the link on the admin tab)

kdid (approved):

PR #1979, opened 2021-03-11
Don't error on tech metadata fields with null value

kdid:
- Does not throw error if a technical metadata field has a null value.
- 

Screen Shot 2021-03-11 at 1 01 23 PM


mbklein (approved):

davidschober:
I'm still seeing this on staging with the above works. Does this fix just fix now forward?

PR #1982, opened 2021-03-11
Implement Meadow.Seed.Export and Import

mbklein:
## Description

```
$ mix help meadow.seed.export

                             mix meadow.seed.export

Export images and data from Meadow to an S3 bucket.

## Command line options

  • --ingest_sheets - how many ingest sheets (with associated data) to
    export (default: 0)
  • --works - how many non-ingest-sheet works (with associated data) to
    export (default: 0)
  • --bucket - target S3 bucket (default: the configured Meadow uploads
    bucket)
  • --prefix - (required) S3 prefix for exported assets
  • --skip-assets - output data only, no preservation or pyramid files
    (default: false)
  • --threads - how many uploads to perform at once (default: 1)

$ mix help meadow.seed.import

                             mix meadow.seed.import

Import images and data from Meadow from an S3 bucket or a directory on disk.

## Command line options

  • --prefix - (required) the location of the export, either:
    • an s3:// URI pointing to an export directory hosted in S3
    • a file:// URI pointing to an export directory on the local disk
    • a non-URI path to an export directory on the local disk

  • --threads - how many uploads to perform at once (default: 1)
```

### On the Staging Console

To export 1 ingest sheet and 10 unaffiliated works (plus all linked projects, rows, works, file sets, etc.) to s3://meadow-s-uploads/path/to/my/export/:

```
$ mix meadow.seed.export --prefix path/to/my/export --ingest-sheets 1 --works 10
```

(Note: export uses the upload bucket by default, but can be overridden with the `--bucket` command line option)

There is a sample export (2 ingest sheets + 20 standalone works) available at `s3://meadow-s-uploads/seed-test/` that you can use to test importing into your dev environment.

To import into dev (**Note: This will wipe out any existing data in your dev environment!**):

```
$ aws-adfs login --profile staging
$ aws s3 sync s3://meadow-s-uploads/path/to/my/export/ /path/to/local/export/directory
$ mix meadow.reset
$ mix meadow.seed.import --prefix /path/to/local/export/directory
```

To import into staging (**Note: You must [wipe staging data](http://docs.rdc.library.northwestern.edu/2._Developer_Guides/Meadow/Migration/#wipe-meadow-staging-data) before importing**)

```
$ mix meadow.seed.import --prefix s3://meadow-s-uploads/path/to/my/export/
```

## Notes

* The import task will roll back your database to the same version present in the source system at the time of the export, and then migrate it back forward after the import is complete. You should make sure that the branch you're importing on is a fast-forward or clean merge ahead of the revision where the export was run. When in doubt, rebase on `deploy/staging`
* As noted above, import requires an _empty_ instance (`down -v` on dev, `mix meadow.reset` on staging). If the import database already has data in it, you will probably encounter unique constraint violations and the import will fail.
* Because of how S3 configuration works, it is not possible to export from production or staging to S3 and then import directly into dev from S3. You will have to sync the export folder to the local disk and then import from there. However, with a seed bucket shared between staging and production, it should be easy to move data quickly from one to the other.


adamjarling (approved):
Works great.

PR #1993, opened 2021-03-16
Index metadata update job ID with work

mbklein:


kdid:
This looks good as far as the backend work is concerned. The issue has a done looks like "CSV Metadata ID filter added to CSV Dashboard".  Were you going to do that? It be pretty easy to add (just copy the batch dashboard implementation I think).

PR #1994, opened 2021-03-16
Change "Undelabelined" to "Undetermined"

mbklein:
The coded term seed file was the only instance of the search/replace error. I ended up writing a migration to change the label in the `coded_terms` table and in every work it's already been used in, which will also trigger a reindex on the next sync.

### DIFF
--- a/priv/repo/migrations/20210316004128_change_undelabelined_to_undetermined.exs
+++ b/priv/repo/migrations/20210316004128_change_undelabelined_to_undetermined.exs
@@ -1,0 +1,36 @@
+defmodule Meadow.Repo.Migrations.ChangeUndelabelinedToUndetermined do
+  use Ecto.Migration
+  alias Meadow.Data.CodedTerms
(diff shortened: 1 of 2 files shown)


bmquinn (approved):
Looks good 👍 

```
15:11:28.102 [debug] QUERY OK db=2.2ms
UPDATE coded_terms SET label = $1 WHERE scheme = $2 AND id = $3 ["Copyright Undetermined", "rights_statement", "http://rightsstatements.org/vocab/UND/1.0/"]

15:11:28.110 [debug] QUERY OK db=5.6ms
  UPDATE works
  SET descriptive_metadata = jsonb_set(descriptive_metadata, '{rights_statement,label}', $1)
  WHERE descriptive_metadata->'rights_statement'->>'id' = $2
 ["Copyright Undetermined", "http://rightsstatements.org/vocab/UND/1.0/"]
 ```

I think this is a good pattern for migrations going forward, too. Maybe worth discussing as a team.

PR #1998, opened 2021-03-16
Preservation Check

kdid:
Generates a preservation check "report" (csv) file and uploads it to the preservation check bucket. Cycles through all works in system's file sets and checks if the preservation file exists at the file set's `metadata.location`. (will also fail if the `metadata.location` value does not refer to the preservation bucket). Also checks if the pyramid is found at the pairtreed file_set_id path in the pyramid bucket. Stores information about the preservation checks in the database so that they can be exposed and downloaded in the front end.  

- `Meadow.Data.PreservationCheckWriter` takes a filename, generates a report and uploads it to the s3 bucket
 - New bucket for `preservation_checks` - Terraform not applied
 - New schema for backend for preservation check dashboard: `Meadow.Data.Schemas.PreservationCheck` and associated migration.
 - `Meadow.Data.PreservationChecks` API for interacting with `preservation_check` data
 - ~~`Meadow.PreservationCheckDriver`: IntervalTask that runs preservation checks if needed.~~
 - GraphQL query to get preservation check data
 - updates presigned url query to handle preservation checks
 - Exposes Preservation Check dashboard in UI
 - configure `quantum` package to run preservation checks on schedule (set to every 10 min on dev and every day at 2am on prod)
 - Minor version bump for database change

Screen Shot 2021-03-24 at 4 16 38 AM


assets/js/components/Dashboards/PreservationChecks/List.jsx
```diff
@@ -0,0 +1,117 @@
...
+      statusClass = "is-danger";
+      break;
+    case "ACTIVE":
+      statusClass = "is-warning";
+      break;
+    default:
+      break;
+  }
+
+  return <span className={`tag ${statusClass} is-light`}>{status}</span>;
```
adamjarling, line 30: We now have a `Tag` component you can import:  https://nulib.github.io/admin-react-components/styleguide/#tag

assets/js/components/Dashboards/dashboards.gql.js
```diff
@@ -80,6 +80,20 @@ export const GET_BATCHES = gql`
   }
 `;
 
+export const GET_PRESERVATION_CHECKS = gql`
+  query preservationChecks {
+    preservationChecks {
+      id
```
adamjarling, line 86: Can we alphabetize these (except for `id` if you want to keep that at the top)?

assets/js/components/Dashboards/dashboards.gql.mock.js
```diff
@@ -77,6 +78,31 @@ export const mockGetBatchesResults = [
   },
 ];
 
+export const mockPreservationChecks = [
+  {
+    id: "7c171c70-1f1a-4db8-8dcb-07c2c0003e10",
```
adamjarling, line 83: Same here, just to make code reading fast, if we can alphabetize that'd be great.

adamjarling (changes requested):
This all looks great.  Few small cosmetic requests...

mbklein (changes requested):
I see that `pyramid_exists` is `N/A` for file sets with a role of `P`, but I think it would make sense to also leave the `pyramid_location` value empty for those. I was confused by what `N/A` meant until I saw they were all `P` files.

I'd also like to see consistent naming between `exists_in_preservation` and `pyramid_exists` (e.g., `preservation_exists`) but I'm not super bothered about it.

lib/meadow/data/preservation_check_writer.ex
```diff
@@ -171,11 +171,13 @@ defmodule Meadow.Data.PreservationCheckWriter do
     end
   end
 
-  defp pyramid_uri_for(file_set_id) do
+  defp pyramid_uri_for(%{role: %{id: "P"}}), do: ""
```
mbklein, line 174: 👏 👏

lib/meadow/data/preservation_check_writer.ex
```diff
@@ -17,7 +17,7 @@ defmodule Meadow.Data.PreservationCheckWriter do
     "file_set_role",
     "sha256",
     "sha1",
-    "preservation_location",
+    "preservation_exists",
```
mbklein, line 20: This looks like the wrong field? This will result in `preservation_exists` and `exists_in_preservation` instead of `preservation_location` and `preservation_exists`.

mbklein (changes requested):

lib/meadow/data/preservation_check_writer.ex
mbklein: (This one isn't a change request. I just love seeing pattern-match solutions to issues like this.)

PR #1999, opened 2021-03-17
Add search to NUL Authorities dashboard screen

adamjarling:
This is now wired up with tests passing.

image

### DIFF
--- a/assets/js/components/Dashboards/Csv/List.jsx
+++ b/assets/js/components/Dashboards/Csv/List.jsx
@@ -68,7 +68,7 @@ function DashboardsCsvList(props) {
     <div data-testid="csv-list">
       <UISearchBarRow isCentered>
         <UIFormInput
-          placeholder="Search"
+          placeholder="Filter"
           name="csvSearch"
           label="Filter CSV imports"
           onChange={handleFilterChange}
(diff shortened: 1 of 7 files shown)


kdid:
It works great for the initial search. The behavior was a little weird when I searched for a term and then did an update on it:

I searched "dogs", got 3 results, clicked edit on one of them and added/changed the hint. 
After save (I got a success message) but did not see the updated hint in the UI, and still was only shown the 3 results for dogs but "dogs" was no longer present in the search box. 

I think there's a couple different ways this flow could work...

adamjarling:
Ah, yea good call. I'll try to force the refresh of data and see if I can retain the current search term.

adamjarling:
@kdid This is cleaned up now so after you edit an authority record, it saves your search and keeps the filter.

kdid:
The search term stays in the search box now but the page doesn't show the updated entry. Should it? I'm not trying to make this task impossible but as a user I would be wondering if my edit worked or not. If that's too hard to implement I'm willing to approve.

adamjarling:
@kdid So this was odd, it was working with some test data I previously was using, but not any new data.  Guessing my test data from yesterday was one of the first 100 records returned, so it was triggering the `useEffect` hook to update values.  This latest commit explicitly calls the filtering function from both an update, and a delete.  I tested with a few records and seems to work for me, but if you can put it through the ringer I appreciate it.

kdid (approved):

kdid:
Yep. Looks good.

PR #2029, opened 2021-03-22
Mobile cleanup

adamjarling:
Not perfect, but all pages are mostly viewable now and I think a decent start.

This PR also establishes a pattern of simplifying the `import` of common `UI` components, and creates some new UI components which handle desktop/mobile display. 

image

image

### DIFF
--- a/assets/js/components/Charts/GoogleAnalytics.jsx
+++ b/assets/js/components/Charts/GoogleAnalytics.jsx
@@ -2,21 +2,25 @@ import React from "react";
 import PropTypes from "prop-types";
 import { Link } from "react-router-dom";
 import IconChart from "@js/components/Icon/Chart";
+import UIActionHeadline from "@js/components/UI/ActionHeadline";
 
 function ChartsGoogleAnalytics(props) {
   return (
     <div>
-      <div className="is-flex is-justify-content-space-between">
-        <h3 className="subtitle is-3">
-          Digital Collections traffic last 30 days
-        </h3>
-        <Link to="/dashboards/analytics" className="button">
-          <IconChart className="icon" />
-          <span>View GA Dashboard</span>
-        </Link>
-      </div>
+      <UIActionHeadline>
+        <>
+          <h3 className="subtitle is-3">
+            Digital Collections traffic last 30 days
+          </h3>
+          <Link to="/dashboards/analytics" className="button">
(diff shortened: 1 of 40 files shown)


kdid:
Looks good. Here's a couple things I noticed. Not sure if these are maybe things that are out of scope for now:

- Lost hamburger menu on a couple pages (ingest sheet page, work page)
- Full Google Analytics dashboard is **_tiny_**
- Preservation tab table squashed and unreadable (most of the other tables seem to scroll off to the right)

Screen Shot 2021-03-23 at 2 12 11 PM
Screen Shot 2021-03-23 at 2 13 29 PM
Screen Shot 2021-03-23 at 2 10 01 PM
Screen Shot 2021-03-23 at 2 18 24 PM

adamjarling:
@kdid Nice catch on the Ingest Sheets page...I missed that one. 

The other 2 (GA dashboard, and the Preservation table) I was aware of but just skipped.  

@davidschober For the GA Dashboard, since this is iFrame content, not sure how we go about this on our end, except to design the dashboard "mobile-first", then let it unfold out to Desktop?   Or we provide just a button on mobile, which links directly to the Google hosted page?

The Preservation table is going to take some extra work, so I'll create a ticket it for that now.

davidschober:
>  For the GA Dashboard, since this is iFrame content, not sure how we go about this on our end, except to design the dashboard "mobile-first", then let it unfold out to Desktop? 
I'm not sure I know what unfold means in this context. 

> Or we provide just a button on mobile, which links directly to the Google hosted page?
I think that might be enough for this use case. I think we want to optimize for laptop/desktop for these reports otherwise we will spend more cycles than we need to.

kdid:
Ingest sheet page looks better but hamburger menu still missing from work page.

Screen Shot 2021-03-24 at 1 11 31 PM

adamjarling:
@kdid So the hamburger menu likely isn't missing, it's just being pushed to the right by some piece of longer metadata which isn't being "word breaked" in metadata display, in the Tabs section.  

I'd say let's break this into a smaller issue, as the point of this issue was to get like 90% of the way there on mobile (given that it started around 20%)

kdid:
Oh yeah that's cool I didn't realize it was over to the right. 
Approving

kdid (approved):

PR #2078, opened 2021-03-30
Support bg colors for staging and optional toggle switch for dev envi…

adamjarling:
…ronment

#### Staging environment
image

#### Dev environment
image

image

Also added support for anything we want to save in Meadow for the future, in `localStorage` in case user's want to customize UI appearance on things in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Now you can visit [`https://devbox.library.northwestern.edu/`](https://devbox.li
 
 Start the Phoenix with `mix phx.server` or `iex -S mix phx.server` if you want to an interactive shell.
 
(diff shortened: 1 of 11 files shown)


mbklein (approved):

adamjarling:
@kdid Yea, it should be sticking....looking into it.

adamjarling:
@kdid Made another round of updates which simplifies the implementation.  The key takeaways are:

- When the Dev bg color toggle switch is toggled, a page refresh occurs to initiate the updated `localStorage` value.

image

- When Dev bg color is active (and will persist active), you can update your own background color via the `devBg` `localStorage` value in your browser's console.

image

kdid (approved):

PR #2099, opened 2021-04-05
Upload source maps to honeybadger after phx.digest

mbklein:
The minified URLs for Meadow's source maps in Honeybadger don't include the Phoenix digests:

image

(Those should be `vendors~app.bundle-7b5fbdaaf3c862113bdfe4e2fbf33d79.js` and `app.bundle-55fb79a6f44cbbf586da786ab53c5a1e.js` to match what's in staging.)

The `@honeybadger-io/webpack` plugin doesn't know about Phoenix digests, so we have to change things up and upload source maps manually.

This PR also includes a little bit of refactoring to make the CircleCI build config easier to read by moving large, multi-line `run` commands to separate bash scripts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,6 +146,7 @@ jobs:
           image: nulib/meadow
           tag: ${DEPLOY_TAG}
       - run:
+          name: Tag Deps Image
           command: |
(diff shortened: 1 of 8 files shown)


bmquinn (approved):
Really like refactoring the commands to their own bash scripts! 👍

PR #2111, opened 2021-04-07
Update dev bg toggle for staging environment

adamjarling:
Default bg color, but a user can update this in their `localStorage.devBg` property to any color they like.

Also adds either "DEV" or "STAGING" to the header next to the Meadow logo.

image

### DIFF
--- a/assets/js/components/UI/DevBgToggle.jsx
+++ b/assets/js/components/UI/DevBgToggle.jsx
@@ -2,7 +2,7 @@ import React from "react";
 
 // Default color light blue
 // You can update this in your browser's console by setting a new hex value for "devBg" property
-const BG_HEX_VALUE = "#eff5fa";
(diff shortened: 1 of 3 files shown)


mbklein:
Am I correct that `localStorage.devBg` is _not_ supposed to persist between toggles? Like, if I toggle the dev background on (shows blue), then change it to red in my console and reload (shows red), then toggle it off and on again, it should go back to blue?

adamjarling:
@mbklein  Yes, correct.  Toggling the switch either sets a text string value, or an empty string value.    Essentially resetting the color to the default.

adamjarling:
But we can also save the custom bg color if you think that'd be helpful?

mbklein (approved):

mbklein:
We can wait and see if that's something the end users think would be useful. Looks good for now.

PR #2113, opened 2021-04-07
Intercept browser back button event on Search screen, and take user t…

adamjarling:
…o previous screen to avoid a double back click.

This solution is only implemented on the Search screen.  Figure tread lightly manipulating the `history` stack and see if this is a more pleasant UX?

To test this:
1. Go to any Collection or Project that has a "View Works" button. 
2. Click the button
3. After Search screen has loaded, click browser "Back" button.

One click should get you back to the previous screen.


bmquinn (approved):
👍

PR #2114, opened 2021-04-07
Add COMPLETED_ERROR status to ingest sheets

mbklein:
* Update the existing progress trigger to set the completed ingest sheet status to `completed_error` if the sheet encountered any errors during processing
* Add the new `COMPLETED_ERROR` status to the GraphQL API
* Update the front end to display the new status

Current styling for the `COMPLETED_ERROR` status is red (`is-danger`) but can be changed easily if a different presentation would be better.

To test: Submit an ingest sheet pointing to a file that exists but that isn’t a valid image.


bmquinn (approved):
Looks great 👍 
complete_with_errors

adamjarling (approved):

PR #2118, opened 2021-04-08
Refactor hard-coded limit in AuthorityRecords.list_authority_records/1

bmquinn:
… to include a limit argument with a default


mbklein:
This change adds a parameter to an API call, but it's an optional parameter with a sane default, so it's not a breaking change. Does that make this a point update or a minor update?

bmquinn:
Good question, I'd say point update if forced to choose.

adamjarling (approved):

PR #2127, opened 2021-04-09
Update title in page head, and put in protection against Helmet error…

adamjarling:
…s on empty Collection or Work titles


bmquinn (approved):
Awesome, fixes the issue for my untitled works in dev 👍

PR #2150, opened 2021-04-16
Rework Meadow.Error.Filter to use Honeybadger.Filter.Mixin

mbklein:


bmquinn (approved):
Awesome, glad Honeybadger fixed the upstream issue! Simplifies this logic quite a bit 👍

PR #2169, opened 2021-04-26
Update Project Metadata to be single value, on the UI side

adamjarling:
On the UI side, these form inputs are now just regular text inputs.  Before sending data to the API for update mutations, the single value is just getting cast as an `array` data type.  

When the back end is ready to accept and deliver single value data, can do some more clean up on the UI side, and also re-organize Project metadata helper arrays a bit.

image

image


bmquinn (approved):
Works great, 👍

PR #2172, opened 2021-04-27
Append to NULL array is the same as replace

mbklein:
NOTE: To test this properly, you must start with seed data from before the existence of the `cultural_context` field


adamjarling:
@mbklein @bmquinn Still not sure I'm doing this properly, but I haven't re-seeded data since last week.  Same data as when I noticed the bug.

I just tried to append a new Cultural Context value via Batch Edit, and still getting the same error.  Do I need to `devstack down -v` ?

image

mbklein:
Yeah, maybe try with a down -v and re-seed?

adamjarling:
Yea, success.  Looks good on this end.

adamjarling (approved):

PR #2194, opened 2021-05-07
Download filesets on Work Preservation tab

adamjarling:
Cleaned up the Preservation tab table row, including a Bulma `dropdown` component which is now converted to a React `UIDropdown` and `UIDropdownItem` component for our app.

Pulled in `@samvera/image-downloader` package, and added this to the Structure tab, replacing previous "JPG" download button.

Note downloading a TIF is not implemented here, and we should create a separate issue for how to handle that as it'll be a little more involved between packages.

image

image


kdid:
@adamjarling - I'm confused about this. I'm not seeing the download functionality on either the Structure or Preservation tab. Should I be?

kdid (approved):

PR #2219, opened 2021-05-18
Add work_type and work_image fields to ingest sheet and validator

mbklein:


kdid:
@mbklein - it doesn't seem like you've included the Elixir code that supports the functionality? - Bullet point #2 in the issue. (Ex: setting the work image).

mbklein:
Ah, I didn't realize that specific part of the functionality was in scope for this, but it makes sense that it would be. I thought “functionality” just meant making it validate and that actually setting the work properties would be a different task. I'll add it to this one, though.

kdid:
No, we were thinking it was included. We pointed it pretty high and didn't break out a separate issue

mbklein:
There is a test for it, right [here](https://github.com/nulib/meadow/pull/2219/files#diff-6b9d1c701bd7d3d7fcb16b0153cdf1085f6f696ddcfb2d2ee9800436656399a2R29). (Ingest sheet line [here](https://github.com/nulib/meadow/pull/2219/files#diff-0e69111a71a6318694878c27073995dd01e86812514fde6229fc50d0b9568969R4).)

I'll mess with it locally as well.

kdid:
Ok, I think it was my bad. I got it working. It does seem to be allowing a "P" file to be a representative image, which is breaking the work.

kdid (approved):

PR #2242, opened 2021-05-27
Pipeline dispatcher refactor

kdid:
- Change the pipeline so that the actions up to `ExtractMimeType` are common, after that a Dispatcher is initialized which handles routing through the correct processing pipeline based on file set role + mime type. 

Docs PR: https://github.com/nulib/repodev_planning_and_docs/pull/1922

### DIFF
--- a/lib/meadow/ingest/rows.ex
+++ b/lib/meadow/ingest/rows.ex
@@ -23,6 +23,16 @@ defmodule Meadow.Ingest.Rows do
   end
 
   @doc """
+  Gets a row by ingest sheet and file set accession_number
+  """
+  def get_row_by_accession_number(sheet_id, file_set_accession_number) do
+    from(r in Row,
+      where: r.sheet_id == ^sheet_id and r.file_set_accession_number == ^file_set_accession_number
+    )
+    |> Repo.one()
+  end
+
+  @doc """
   Changes the validation state of a Row
   """
   def change_ingest_sheet_row_validation_state(
--- a/lib/meadow/ingest/rows.ex
+++ b/lib/meadow/ingest/rows.ex
@@ -25,7 +25,7 @@ defmodule Meadow.Ingest.Rows do
   @doc """
   Gets a row by ingest sheet and file set accession_number
   """
-  def get_row_by_accession_number(sheet_id, file_set_accession_number) do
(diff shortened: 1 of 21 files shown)


lib/meadow/pipeline/dispatcher.ex
```diff
@@ -0,0 +1,123 @@
...
+    do: @preservation_actions
+
+  def dispatcher_actions(_), do: nil
+
+  def not_my_actions(file_set) do
+    (all_progress_actions() -- @initial_actions) -- dispatcher_actions(file_set)
+  end
+
+  defp dispatch_next_action(file_set, attributes) do
+    last = ActionStates.get_latest_state(file_set.id)
```
mbklein, line 95: You can get the name of the last action (and its completion status) straight out of the `attributes` hash:

```
%{process: "CreatePyramidTiff", role: "A", status: "ok"}
```

(These are the same attributes referenced in `sequins.exs` that sequins uses to create subscriptions.)

I think the advantage to using the `process` attribute here is for when we might want to manually run a file set back through one or more actions. `ActionStates.get_latest_state/1` has a better chance of being confused by the existence of other actions, whereas `process` will tell you exactly what just finished. You just have to use 

```
Module.safe_concat(Meadow.Pipeline.Actions, action_name)
```

to get the fully qualified module name of the action (e.g., to turn the string `"CreatePyramidTiff"` into a reference to the module `Meadow.Pipeline.Actions.CreatePyramidTiff`).

mbklein (changes requested):

lib/meadow/ingest/rows.ex
```diff
@@ -22,6 +22,16 @@ defmodule Meadow.Ingest.Rows do
     |> Repo.one()
   end
 
+  @doc """
+  Gets a row by ingest sheet and file set accession_number
+  """
+  def get_row_by_accession_number(sheet_id, file_set_accession_number) do
```
bmquinn, line 28: Since there's already a `Rows.get_rows_by_work_accession_number/2`, can you name this `Rows.get_rows_by_file_set_accession_number/2` instead? I think a test for this is needed as well since we're expecting only one result, would be good to have that verified in a test.

bmquinn (changes requested):

bmquinn (approved):

mbklein (approved):

PR #2244, opened 2021-05-27
Add structural metadata to file sets

mbklein:
- Create structural metadata field
- Update GraphQL queries, mutations, and types to include structural metadata
- Create database listener to write VTT file to S3


kdid (approved):

bmquinn (commented):
Looks great!

vtt

The only question I had was whether you think we should handle failures in calls to `write_structural_metadata/1` with `Meadow.Error.report/4`. I'm thinking we probably want Honeybadger to shout if this fails so that we can fix things manually.

bmquinn (approved):

PR #2249, opened 2021-05-28
Add Sheets.kick!/1 to manually fix stuck ingest sheets

bmquinn:
- Adds Sheets.kick!/1 which provides the ability to manually fix a "stuck" ingest sheet

### DIFF
--- a/test/meadow/ingest/sheets_test.exs
+++ b/test/meadow/ingest/sheets_test.exs
@@ -3,7 +3,8 @@ defmodule Meadow.Ingest.SheetsTest do
 
   alias Meadow.Data.Works
   alias Meadow.Ingest.Schemas.Sheet
-  alias Meadow.Ingest.Sheets
+  alias Meadow.Ingest.{Progress, Rows, Sheets}
+  alias Meadow.Utils.MapList
 
   describe "ingest_sheets" do
     @valid_attrs %{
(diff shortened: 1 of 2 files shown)


test/meadow/ingest/sheets_test.exs
```diff
@@ -149,4 +150,27 @@ defmodule Meadow.Ingest.SheetsTest do
       assert Sheets.work_count(ingest_sheet) == 30
     end
   end
+
+  describe "fix stuck ingest sheet" do
+    @fixture "test/fixtures/ingest_sheet.csv"
+
+    setup do
```
bmquinn, line 157: This test setup ended up being pretty hard to figure out, any suggestions to improve this are welcome!

kdid:
If we already have a database listener that sends notifications when the status is updated why wouldn't we just update the status directly rather than doing a fake update of a row just to kick the status update? 

https://github.com/nulib/meadow/blob/deploy/staging/lib/meadow/ingest/sheet_notifier.ex

mbklein:
We have a database listener that sends notifications when `ingest_sheets.status` is updated, but the issue being addressed here is that the sheet status is explicitly _not_ being updated because of a race condition. This is an issue with a trigger on the `ingest_progress` table, not the `ingest_sheets` table. Creating a `kick!` function gives us an easier way to intervene when a sheet gets stuck due to the race condition (and possibly eventually mediate it through a listener or an interval task).

kdid:
@mbklein  - I understand the problem, what I don't understand is why this is an easier solution

mbklein:
I don't understand what the alternative is. I see the problem like this: We have a sheet that _appears_ stuck. We can't just update the sheet status, because we don't know _for sure_ that it's stuck, and if it is, we also don't know whether its final status should be `completed` or `completed_error`. We can either inspect the progress rows manually to figure out both of those things, or we can write Elixir code to determine those things and then update the `ingest_sheets`. But we already have PL/pgSQL code written to do all of that – in the `ingest_progress` trigger. Every once in a while, that code fails due to a race condition, but I think it's far easier and more maintainable to keep the code in one place and sometimes manually poke the trigger as a way to say, “hey, I think this sheet is finished, could you please check?” If it really is stuck, the trigger will set it to the correct of the two completed statuses. If not (i.e., some progress rows are still `pending`), no harm done.

Ideally, we would make it so the race condition doesn't exist in the first place. That was the purpose of adding [advisory locking](https://github.com/nulib/meadow/commit/d522808) to the trigger, but it hasn't eliminated the problem. I haven't thought of anything else that wouldn't introduce a performance bottleneck.

bmquinn:
In my view, this was simply a way to codify the fix that we ran twice in production already. So, rather than having to remember how to fix a stuck ingest you can just call this function passing in the sheet without having to recall anything about Progress entries, rows, and statuses.

kdid:
Yeah, I forgot about the two different completed statuses.

kdid (approved):

PR #2252, opened 2021-05-28
Add download shared links CSV file to Search screen; refactor batch b…

adamjarling:
…uttons in Search modal

DRY'd up some code in this component while in there.

Also thanks to @kdid 's keen eye, this PR includes the addition of an inflection package to handle singular/plural labeling in the UI.  Works great.

#### UPDATED
image

image

#### OLD
image

image

### DIFF
--- a/assets/js/components/Search/ActionRow.jsx
+++ b/assets/js/components/Search/ActionRow.jsx
@@ -11,7 +11,17 @@ import {
   IconMinus,
   IconTrashCan,
 } from "@js/components/Icon";
+import SearchFormDownloadWrapper from "@js/components/Search/FormDownloadWrapper";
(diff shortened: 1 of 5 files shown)


kdid:
Screen Shot 2021-06-01 at 7 29 13 AM


Works great, I did notice that there is a less than sign on the button after the "Batch edit X works" button in the select individual mode modal. (not sure if it was caused in this PR). Also, does JS have an inflection package you can use for the plural/singular case? (Not a huge deal as I don't think batch editing one is a main use case....)

adamjarling:
Ah good catch on the typo @kdid .  I'll fix that and see if there's a package to handle singular/plural

kdid (approved):

PR #2228, opened 2021-05-19
Create mediainfo lambda and action

mbklein:
* Create mediainfo lambda
* Create ExtractMediaMetadata action
* Change file set metadata 'exif' to 'technical'
* Add migration to convert exif metadata to generic technical metadata

This PR includes a major version bump to `2.0.0` due to its backwards-incompatible GraphQL changes

PR #2259, opened 2021-06-01
Create TranscodeComplete pipeline action

mbklein:
* Allow front end to accept `audio/*` and `video/*` files
* Update `MediaConvert.Mock` to send an appropriate EventBridge message to the `transcode-complete` queue
* Add correct queue configuration for `create-transcode-job` and `transcode-complete`
* Integrate audio and video types into the dispatcher
* Add a `derivatives` field to the `FileSet` schema that will hold a named map
  * Update `CreatePyramidTiff` action to write the pyramid TIFF location to `derivatives -> pyramid_tiff`
  * For audio and video, write the playlist location to `derivatives -> playlist`
  * Create a mix task to add pyramid TIFF locations for existing file sets to `derivatives` at a later date

### DIFF
--- a/lib/meadow/pipeline/actions/transcode_complete.ex
+++ b/lib/meadow/pipeline/actions/transcode_complete.ex
@@ -1,0 +1,79 @@
+defmodule Meadow.Pipeline.Actions.TranscodeComplete do
+  @moduledoc """
+  Action to handle job state change messages from MediaConvert
+  """
+
+  alias Meadow.Data.{ActionStates, FileSets}
+  alias Meadow.Ingest.{Progress, Rows}
+
+  use Sequins.Pipeline.Action
+  use Meadow.Utils.Logging
+
+  require Logger
+
+  @impl true
+  def process(%{detail_type: "MediaConvert Job State Change"} = message, _attributes) do
+    with {data, attributes} <- parse(message) do
+      process(data, attributes)
+    end
+  end
+
+  def process(%{file_set_id: file_set_id} = data, attributes) do
+    with_log_metadata module: __MODULE__, id: data.file_set_id do
+      with {status, response} <-
+             file_set_id |> FileSets.get_file_set() |> process_mediaconvert_response(data) do
+        update_progress(attributes, status)
+        {status, response, attributes}
+      end
+    end
+  end
+
+  defp parse(message) do
+    {%{
+       file_set_id: message |> get_in([:detail, :user_metadata, :file_set_id]),
+       status: message |> get_in([:detail, :status]),
+       error: message |> get_in([:detail, :error_message]),
+       playlist: message |> extract_playlist()
+     },
+     message
+     |> get_in([:detail, :user_metadata])
+     |> Map.delete(:file_set_id)}
+  end
+
+  defp extract_playlist(message) do
+    case message |> get_in([:detail, :output_group_details]) do
+      [detail | _] ->
+        detail |> get_in([:playlist_file_paths, Access.at(-1)])
+
+      _ ->
+        nil
+    end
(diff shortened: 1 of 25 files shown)


lib/meadow/pipeline/actions/transcode_complete.ex
```diff
@@ -0,0 +1,77 @@
...
+
+  defp process_mediaconvert_response(nil, %{file_set_id: file_set_id}) do
+    Logger.warn(
+      "Marking #{__MODULE__} for #{file_set_id} as error because the file set was not found"
+    )
+
+    {:error, "FileSet #{file_set_id} not found"}
+  end
+
+  defp process_mediaconvert_response(file_set, %{status: "skip"}) do
```
kdid, line 61: The idea of sending a file set through all the actions and skipping the action and marking "n/a" if the file set was of the wrong mime-type, was the idea we had before we settled on the Dispatcher idea. Now the dispatcher is the one responsible for making determinations based on mime-type. 

Now that the Dispatcher is merged, I think this should come out (along with the associated code from the CreateTranscodeJob as well), unless we plan to implement the same pattern for all the actions. It doesn't seem right that this should behave differently than all the other actions and put a different status in the d/b. 

I think it should error like CreatePyramidTiff does if you pass it a text file.

kdid:
The progress doesn't seem to be going to 100% (I'm using the `ingest_sheet.csv` fixture to test in the UI and it sticks at 32%, with a bunch progress entries still set to `pending`.

lib/meadow/pipeline/actions/transcode_complete.ex
mbklein: Agreed. I made a similar comment on the CreateTranscodeJob PR and forgot to look back at it when writing this one. I’ll remove it.

mbklein:
Weird. It went through for me yesterday but possibly something got borked in a rebase. I’ll check it out this morning.

mbklein:
That issue was completely covered (as far as its “Done Looks Like” criteria) by #2228, but the code got unlinked from the Sequins pipeline config by other PRs. I don't see a problem connecting this PR to it as well if it would be useful.

bmquinn (approved):

lib/meadow/pipeline/actions/transcode_complete.ex
mbklein: I've removed both the mime type test in `CreateTranscodeJob` and the `skip` response handling from `TranscodeComplete`.

kdid:
Well, I was asking because of what we talked about at standup - the tech metadata action for A/V in the pipeline. I thought you said you were also adding that work to this PR. And that other issue was in "In Progress" so I thought maybe that was the issue.

mbklein:
I just tried to connect this PR to that issue, but ZenHub won't let me.

image

kdid:
No worries.

Docs need to be updated too: http://docs.rdc.library.northwestern.edu/2._Developer_Guides/Meadow/Batch-Ingest-and-Ingest-Pipeline/

obsolete since yesterday

mbklein:
Added nulib/repodev_planning_and_docs#1935 to address the documentation issue.

priv/repo/migrations/20210602143423_fix_new_old_bug_in_ingest_sheet_update_trigger.exs
```diff
@@ -0,0 +1,88 @@
+defmodule Meadow.Repo.Migrations.FixNewOldBugInIngestSheetUpdateTrigger do
```
kdid, line 1: I'm thinking separate PR for this?

PR #2263, opened 2021-06-02
Add support for audio and video work types, but rendering a hardcoded…

adamjarling:
… media file (for now). This work supports rendering (hardcoded video still) of a non-image Work Type needed for the demo.  And for now, just opens an Aux image file in a new browser tab.

image


kdid:
This is not to be reviewed/merged, correct?

adamjarling:
Changes reflected in https://github.com/nulib/meadow/pull/2275

PR #2268, opened 2021-06-03
Only run the ingest_progress trigger on update, not delete

mbklein:


kdid:
@mbklein are we still merging this? Did you say it will have no effect?

mbklein:
It will have an effect in preventing cascade errors on delete. It just won’t fix the timeouts on create that I thought it might help (because it already wasn’t firing on create).

kdid (approved):

PR #2272, opened 2021-06-03
Add streaming to dev environment via MediaConvert.Mock

mbklein:
## Description

This PR adds the ability to stream audio/video content directly from minio in the developer environment. Instead of sending the media to MediaConvert for transcoding, it simply copies the source file to the streaming bucket and returns the location of the file as the playlist. Minio can serve them to the front end via HTTP.

### Testing

1. Create a work with an audio file set and a video file set.
2. Save the following to a file called `streaming_test.html` somewhere outside your Meadow directory.

    ```html
    <html>
      <head>
        <title>Streaming Test</title>
      </head>
      <body>
        <p>
          <video controls>
            <source src=""/>
          </video>
        </p>
        <p>
          <audio controls>
            <source src=""/>
          </audio>
        </p>
      </body>
    </html>
    ```

3. Request the work in graphiql and get its file sets' `streamingUrl` properties.
4. Paste the URL for the video into the video source and the audio into the audio source above.
5. Load `streaming_test.html` in a browser.

You should have fully functioning video and audio players in the browser window.

image

### Notes
* In order to work with different media types, the front end either needs to match the `audio`/`video` tag's `type` attribute to the file extension of the URL, or just leave out the `type` attribute altogether and let the browser figure it out.
* In order to make streaming work with a Minio bucket, the `Config.streaming_host()` function needed to be replaced with `Config.streaming_url()` to return the entire URL stem. All downstream functions have been updated to account for this, but the Terraform change in this PR will have to be applied after merge.


kdid (approved):

kdid:
I approved, but so question just in terms of front end development. Here the streaming url is the actual audio/video file and in staging/prod it's the m3u8 file. That's something the front end is going to need to be aware of during development and handle  the environment cases differently?

mbklein:
Nope. Given the source URL, the browser will attempt to play it, whether it’s an mp4 or m3u8 or any other valid audio/video file type. If the front end chooses to include the `type` attribute, it will have to use the right one based on the file type (which will always be `video/x-mpegURL` or `audio/mp4` in staging and production, but will vary in dev). But leaving off the `type` and letting the browser figure it out is OK too. The only time `type` is absolutely required is when the URL’s extension doesn’t match its actual content type.

mbklein:
But yes, a show and tell is a good plan.

kdid:
I attached it to this issue: #1876 as it solves all the bullet points

bmquinn (approved):

PR #2274, opened 2021-06-04
Fix mime-type extraction through UI upload

kdid:
- We need an extension on file sets uploaded through the UI so that mime-type identification can work.

This PR:
- changes presignedUrl API for file sets so that it uses the original file extension for the presigned url
- changes DropZone for file set upload to accept all file types (for now)
- changes `getPresignedUrl` on front end from `useQuery` to `useLazyQuery` (since if it fires when the component loads it won't know the file extension)

To test: upload a .json, .csv etc file set in the UI. (Supplemental)


adamjarling (approved):
Looks good and everything seems to work fine adding a CSV file.

PR #2275, opened 2021-06-04
Fileset UI updates

adamjarling:
Disables the player for test Video and Audio Work Type Works, and some other UI updates related to filesets.

image


kdid:
I think the error after upload is from here: https://github.com/nulib/meadow/blob/fileset-ui-updates/assets/js/components/Work/Fileset/ActionButtons/Access.jsx#L49

Screen Shot 2021-06-04 at 2 57 49 PM


Mime-type is something that is determined in the pipeline so wouldn't be available right after the form is submitted

adamjarling:
> I think the error after upload is from here: https://github.com/nulib/meadow/blob/fileset-ui-updates/assets/js/components/Work/Fileset/ActionButtons/Access.jsx#L49
> 
> Screen Shot 2021-06-04 at 2 57 49 PM
> 
> Mime-type is something that is determined in the pipeline so wouldn't be available right after the form is submitted

Yea, you're right.  I have that locally but forgot to commit it.  Pushing up now.

kdid (approved):

PR #2302, opened 2021-06-14
Use proper datacite.resourcetype for audio/video works

mbklein:
* Validate datacite.resourcetype in mock ARK server and tests

### DIFF
--- a/test/meadow/data/works_test.exs
+++ b/test/meadow/data/works_test.exs
@@ -465,4 +465,24 @@ defmodule Meadow.Data.WorksTest do
       end)
     end
   end
+
+  describe "mint_ark/1" do
+    setup %{work_type: work_type} do
+      {:ok, work: work_fixture(%{work_type: %{id: work_type, scheme: "work_type"}})}
+    end
+
+    seed_values("coded_terms/work_type")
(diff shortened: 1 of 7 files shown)


test/meadow/data/works_test.exs
```diff
@@ -465,4 +465,24 @@ defmodule Meadow.Data.WorksTest do
       end)
     end
   end
+
+  describe "mint_ark/1" do
+    setup %{work_type: work_type} do
+      {:ok, work: work_fixture(%{work_type: %{id: work_type, scheme: "work_type"}})}
+    end
+
+    seed_values("coded_terms/work_type")
```
mbklein, line 474: This bit reads the valid `work_type`s from the JSON seed file and creates a separate test per work type. There are two reasons for doing it this way (as opposed to iterating over `CodedTerms.list_coded_terms(:work_type)`:
1. Easier to read the output in case something fails (because the work type will be in the test name)
2. Can't iterate over the `CodedTerms` table outside the test block because the sandbox DB connection isn't checked out yet at that point

bmquinn (approved):
Looks good!

Screen Shot 2021-06-14 at 2 12 50 PM

PR #2331, opened 2021-06-24
Update language surrounding bulk operations

kdid:
Open to other suggestions but this made sense to me:

Screen Shot 2021-06-24 at 2 20 24 PM
Screen Shot 2021-06-24 at 2 20 32 PM
Screen Shot 2021-06-24 at 2 20 49 PM


adamjarling:
@kdid What if we include a verb in the button?  "View bulk actions (all works)"?   Also a button's capitalization should follow how the buttons are handled here in Github too (like a sentence, first word capitalized then lowercase).

kdid:
updated. Sorry I forgot and already squashed the commit.

PR #2344, opened 2021-06-29
Update parsing and rendering of technical metadata for Image and Medi…

adamjarling:
…a work types

image

### DIFF
--- a/assets/js/components/UI/MediaPlayer/MediaPlayer.jsx
+++ b/assets/js/components/UI/MediaPlayer/MediaPlayer.jsx
@@ -3,7 +3,12 @@ import PropTypes from "prop-types";
 import MediaPlayerNav from "@js/components/UI/MediaPlayer/Nav";
 import getVttFile from "@js/services/get-vtt-file";
 
-function MediaPlayer({ sources = [], tracks = [], ...restProps }) {
+function MediaPlayer({
+  sources = [],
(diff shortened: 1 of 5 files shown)


bmquinn (commented):
Technical metadata looks great!

Screen Shot 2021-06-29 at 4 02 17 PM

I noticed a couple small things when I was testing out the branch:

Got an error when uploading a video in the UI that went away after a refresh:
Screen Shot 2021-06-29 at 4 00 23 PM

Noticed an issue in the preservation tab when the browser wasn't full width:
Screen Shot 2021-06-29 at 4 01 23 PM

adamjarling:
Ah, that error message is still sneaking through eh?   I think I fix it, then make some other adjustments which probably knocks it loose.  I'll fix that. 

The table display issue I think should be another issue itself.   Trying to cram all that data into a table for responsive just doesn't seem practical with real data.  I'll spin off an issue and attach the screenshot.  I've been noticing it too.

adamjarling:
@bmquinn Pushing up some changes now which address the error.  It was happening when creating a File Set for a Work that previously had 0 File Sets.   Put a little helper notification message in for now, and handling the conditions.   Maybe later we can work on some status messaging to the UI (if even necessary).

image

bmquinn (approved):
Looks great! 👍

PR #2353, opened 2021-07-06
Rewrite row-level dependency/notification triggers as statement-level triggers

mbklein:
This PR replaces all existing dependency triggers and notification triggers with statement-level equivalents.

### Details

* For notification triggers, instead of firing for every operation on each row, the trigger fires once at the end of the whole operation and sends notifications for each _distinct_ affected row in batches of 100.
* For dependency triggers, instead of updating the relevant parent/child `updated_at` one at a time for each affected row, the trigger uses a single `UPDATE ... WHERE` statement to do them all at once.
* The `DatabaseNotification` module has been updated to handle notifications in batches.
* Migrations run correctly in both directions
* Only one test required a very minor change to its setup code, demonstrating interface compatibility

### Testing

* Make sure you can migrate
* Make sure IIIF manifests and cascading index changes still work
* Make sure nothing else blows up
* Manually check batch/CSV update functionality

### Benchmarks

* Dev database was seeded with 145,927 works and 221,138 file sets from production
* In a SQL session, I ran a series of file set updates. First I updated the description on 100 file sets:

    ```
    update file_sets 
    set core_metadata = jsonb_set(core_metadata, '{description}', '"Update 100 Descriptions"', updated_at = now())
    WHERE id in (SELECT id FROM file_sets LIMIT 100);
    ```

    ran quickly using both `EACH ROW` and `EACH STATEMENT` triggers. But while

    ```
    update file_sets 
    set core_metadata = jsonb_set(core_metadata, '{description}', '"Update All Descriptions"', updated_at = now());
    ```

    never completed using `EACH ROW` (I waited for a half hour), it only took 30 seconds using the `EACH STATEMENT` trigger.

### Notes

The trigger generation code is _extra_ complicated because of the need to keep both the `EACH ROW` and `EACH STATEMENT` logic around (for migrating both up and down). We should consider whether it's worth refactoring migrations again once everything is updated on staging and production so we don't have to deal with all the legacy/backout code. This would require a manual update of the `schema_migrations` table to make the DB recognize the refactored migrations.

### DIFF
--- a/lib/meadow/database_notification.ex
+++ b/lib/meadow/database_notification.ex
@@ -42,7 +42,7 @@ defmodule Meadow.DatabaseNotification do
 
     def up do
       # to be notified of all changes
-      create_notification_trigger(:my_table)
+      create_notification_trigger(:my_table, :all)
 
       # to be notified of changes to specific fields
       create_notification_trigger(:my_table, ["specific", "fields"])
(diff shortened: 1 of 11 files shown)


kdid:
@mbklein - I'm willing to approve it, as long as you have a good plan to really validate everything on staging? 

I do wonder in the long term if there may be a better way we can do all this (not now).

mbklein:
The plan is to snapshot the database before merge, then inspect the database (using psequel or Valentina) after migration, and run some interactive and batch updates on staging to make sure the triggers still function as expected.

kdid (approved):

bmquinn (approved):

PR #2359, opened 2021-07-09
Support adding and editing WebVTT text via form for a media File Set …

adamjarling:
…in the UI

Tried to keep the code contained as much as possible in the `assets/js/components/Work/Tabs/Structure/WebVTTModal.jsx` component, for simplicity.    There's a real-time validator of WebVTT text which will complain if `textarea` content is invalid WebVTT and block form posts.

image

image

image

### DIFF
--- a/assets/js/components/UI/IconText.jsx
+++ b/assets/js/components/UI/IconText.jsx
@@ -18,7 +18,6 @@ function UIIconText({ children, icon, isCentered }) {
 UIIconText.propTypes = {
   children: PropTypes.node,
   icon: PropTypes.node,
-
   isCentered: PropTypes.bool,
 };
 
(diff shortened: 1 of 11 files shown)


kdid:
Overall functionality seems great. Couple things I noticed:

1. When you copy paste in content (which I imagine they would do) it says it's invalid - until you edit it. I think they could ultimately work around this if this is just the best we can do. But is it possible to fix this behavior?

2. The content doesn't refresh until you refresh the page after save. This is just a database save so the updated content should be coming back in the graphQL response. Is it possible to update the cache or whatever?

3. To me it seems like it would be more intuitive if if there is not existing content for structural metadata, the button says "Add structure", but if there was it toggled to "Edit structure" or "Replace structure". If that's an easy check maybe an easy win but if it's hard I'd say separate issue....

adamjarling:
@kdid Yea, I'll fix the stuff above.  The copy thing is a little annoying so will listen for that event.   I built it so the Work query should refetch after the mutation processes, and it was working for me locally.   Maybe I knocked something loose on it.

adamjarling:
@kdid Just pushed up some changes which should fix the issues.  For now I renamed the button "Edit structure (vtt)"...that ok?

kdid (approved):

PR #2365, opened 2021-07-13
Update MediaConvert settings to transcode audio and video to HLS

mbklein:
This PR changes Meadow's audio and video transcoding templates to use HLS instead of CMAF for better compatibility with `video.js` and our range of supported browsers. It uses three built-in MediaConvert presets for video (1080p, 720p, and 540p at reasonable bitrates), but since there are no audio-only system presets, I had to copy the settings we use for AVR over from Elastic Transcoder. The only problem with this is that Terraform's AWS module does not currently support MediaConvert presets, so I had to create them manually (which I've done on both staging and production). nulib/repodev_planning_and_docs#2028 covers the terraforming of those resources.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -63,6 +63,19 @@ config :meadow, :lambda,
   mime_type: {:local, {"nodejs/mime-type/index.js", "handler"}},
   tiff: {:local, {"nodejs/pyramid-tiff/index.js", "handler"}}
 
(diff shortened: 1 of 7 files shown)


mbklein:
I've confirmed that these templates render correctly and produce the desired compatible HLS streams.

kdid (approved):

PR #2386, opened 2021-07-23
Update UI layout title and primary info area for consistency; updated…

adamjarling:
… Work screen and components for consistency

image

image


kdid:
Looks good. The collection screen looked like this (below). Is that how it should look right now?
 (The ..."fill this in..." part)

Screen Shot 2021-07-23 at 12 33 20 PM

adamjarling:
@kdid I did purposely leave that in there.  Better to suppress is?

kdid:
I think so? I mean we don't want it to go to prod like that I don't think.

kdid (approved):

PR #2390, opened 2021-07-26
HLS.js package to play videos in Chrome

adamjarling:
This PR implements the hls.js package which we bind to HTML `video` element to support playback in Chromium browsers.


bmquinn (approved):
Works great in my dev environment! Tested with Chrome and Safari.

mbklein (approved):

PR #2391, opened 2021-07-26
Backend for poster API and generation

kdid:
### TODO

- ~~cleanup poster url artifacts~~
- ~~remove/replace virtual field for representative image on work with stored url~~
- ~~Front end representative Image url~~
- ~~ActionStates?~~
- Audio?
- ~~preservation report - Video access files error looking for pyramid.~~
- ~~dependency trigger? - will indexing happen when needed now that derivatives need to actually be present to form the urls?.~~
- ~~ffmpeg doesn't extract frames for type .mov (local problem only). At least document and let team know.~~
- ~~poster/path resolves in iiif server~~
  - ~~node express app~~ https://github.com/nulib/iiif-express/pull/6
  - ~~stack-s-iiif lambda staging~~
  - ~~stack-p-iiif lambda prod~~
 - lambda container reuse (error message from previous invocation showing)
   -   to cause an error try to create a poster from a .mov on staging
 - ~~Terraform the ffmpeg layer in frame extractor lambda~~

<hr>

### Overview

- API for generating a poster image from an offset in milliseconds.
```
mutation{
  updateFileSet(id: "1bf08f50-d952-4c67-a1c0-2304a953a5ff",
    structuralMetadata: {type: POSTER_OFFSET, value: "700"}
  ){
    structuralMetadata{
      type
      value
    }
  }
}
```
- Work is performed by a new action `GeneratePosterImage` that invokes a new lambda `frame-extractor`
- Poster assets/derivatives are stored in `<pyramids-bucket>/posters/fi/le/se/tp/ai/rt/re/e-poster.jpg`
- Poster offset is stored in structural metadata
```
"structuralMetadata": {
        "type": "POSTER_OFFSET",
        "value": "900"
      }
```
- poster location is stored in the derivatives map
```
%{derivatives: %{"poster" => "s3://dev-pyramids/posters/file-set-id-pairtree/ee-poster.jpg"}}
```
- `representativeImageUrl` is indexed on the FileSet and exposed through the GraphQL api
- `representativeImage` for a work can either contain a pyramid url or a poster url 
-  add `ffmpeg` to CircleCI build


mbklein:
This looks great and is working really well. I just have one question. `structural_metadata` is a single-valued field with a `type` and a `value`. How can I set both a poster image _and_ WEBVTT structure on the same fileset? Writing one overwrites the other. Overwriting the poster offset once the actual poster is generated isn't a big deal, but I don't think updating the poster should wipe out any existing VTT.

mbklein:
Poster extraction doesn't seem to be working on my machine. No matter what video I use or what value I put in for poster offset, I get the following in the log:

```
module=Meadow.Pipeline.Actions.GeneratePosterImage id=f528a1e0-0556-481a-b34b-0b64856c8892 [info]  Generating poster image for FileSet f528a1e0-0556-481a-b34b-0b64856c8892, with playlist: s3://dev-streaming/f5/28/a1/e0/-0/55/6-/48/1a/-b/34/b-/0b/64/85/6c/88/92/bd3f24e1c767266ecd00a2bca297f79fa96d3309a334e5b950b7d6d2839a1b0c.m4v and offset: 3500
module=Meadow.Pipeline.Actions.GeneratePosterImage id=f528a1e0-0556-481a-b34b-0b64856c8892 [error] InvalidParameterType: Expected params.Body to be a string, Buffer, Stream, Blob, or typed array object
module=Meadow.Pipeline.Actions.GeneratePosterImage id=f528a1e0-0556-481a-b34b-0b64856c8892 [error] Error from lambda: s3://dev-pyramids/posters/f5/28/a1/e0/-0/55/6-/48/1a/-b/34/b-/0b/64/85/6c/88/92-poster.tif
```

That error is coming from the [uploadToS3](https://github.com/nulib/meadow/blob/2029-poster-rebased/priv/nodejs/frame-extractor/index.js#L151) function, so it's clear that the lambda is firing and it's getting to the point where it tries to upload to S3 (though with `null` data). But I'm confused that I'm not seeing either the [ffmpeg error message](https://github.com/nulib/meadow/blob/2029-poster-rebased/priv/nodejs/frame-extractor/index.js#L83) or the [transcoding succeeded message](https://github.com/nulib/meadow/blob/2029-poster-rebased/priv/nodejs/frame-extractor/index.js#L86) in my log.

kdid:
@mbklein - I just did a sanity check and it's still working for me. My ffmpeg output looks identical to yours. Just to confirm:
- You did an assets.install in some manner to get the `fluent-ffmpeg` package? 


The error you're getting is the same one I get if I try with a `.mov` file, which you can't extract a frame from. (obviously we should be handling the error better - I'll make a note). We can troubleshoot Monday when you're back.

mbklein (approved):
Approving based on the fact that I was able to replace the local “bare file” video with an `.m3u8` and its associated `.ts` files and do a successful poster update. So it does what it needs to do to work on staging, even if we might have to do a little extra to figure out what to do (or what kind of videos to use) to make it work in dev environments.

PR #2496, opened 2021-09-09
Support setting poster image , first draft, for Access videos

adamjarling:
This PR wires up setting a poster image to the UI for a Fileset Access video.   The user can scrub anywhere in the video, then click the button to set a poster image, which will send an async update.  Since this is not an immediate operation on the backend, for now the user will have to refresh their browser tab to see changes.  There is a note in the confirmation message which pops up letting the user know.   If we want to automate this, we could create an additional ticket.

Also updated the default video placeholder image and poster thumbnails to be 4:3 aspect ratio, which seems to be common with older legacy videos  as opposed to 16:9 HD.

image


bmquinn (approved):
Works great 👍

Screen Shot 2021-09-09 at 4 38 51 PM

PR #2527, opened 2021-09-20
Index rank for file sets

kdid:
- index rank for file sets

### DIFF
--- a/test/meadow/data/indexer_test.exs
+++ b/test/meadow/data/indexer_test.exs
@@ -270,7 +270,7 @@ defmodule Meadow.Data.IndexerTest do
       assert doc |> get_in(["model", "application"]) == "Meadow"
       assert doc |> get_in(["model", "name"]) == "Image"
       assert doc |> get_in(["fileSets"]) |> length == 2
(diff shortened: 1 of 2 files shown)


test/meadow/data/indexer_test.exs
```diff
@@ -369,6 +369,43 @@ defmodule Meadow.Data.IndexerTest do
     end
   end
 
+  describe "file set position encoding" do
```
bmquinn, line 372: I'm going to surprise myself by suggesting that you can remove this `describe` block, because the `rank` setting code is covered elsewhere I think we're ok with the new assertion above.

kdid: @bmquinn you mean just put this test in the describe block above? Or just remove the describe and have the test free-floating?

bmquinn: I was thinking just yank the whole describe block and leave the `assert doc |> get_in(["fileSets"]) |> List.first() |> map_size() == 5`. My main thing is that I'm confident that rank is being set correctly, and we don't have to worry about testing whether one rank is greater or less than another since they're always just integers.

bmquinn (approved):

PR #2683, opened 2021-11-12
Extract digest generation from pipeline to S3 (or Minio) event

mbklein:
# Summary 

The checksum generation was taking too long to complete (1 hour +) for super large video files (50G and up) as a step in our ingest pipeline. We decided to extract checksum generation from the pipeline and use AWS's [Serverless Fixity for Digital Preservation Compliance](https://docs.aws.amazon.com/solutions/latest/serverless-fixity-for-digital-preservation-compliance/welcome.html) step function triggered by an upload to the uploads or ingest buckets. For now, we are generating an `md5` checksum (staging/prod)

# Specific Changes in this PR

- Adds auto-checksum functionality to ingest and upload minio buckets in dev and test environments
- Update the `meadow.buckets.create` task to configure bucket notifications
- Add configuration info to `dev.exs` and `test.exs` to support the above 
- Terraform for Cloud Trail and Event Bridge to invoke the step function
- Refactor `GenerateFileSetDigests` pipeline action to use the checksums stored in the object tags rather than calculate them
- Use the Elixir `WaitForIt` package where appropriate to wait for checksums to be generated before kicking off ingest
- Update GraphQL API to return map of all checksums
- Update Preservation tab's checksum copy functionality to accommodate new API
- Update `Validator` to check for object tags and report error (we consider this to be the easiest first iteration - not necessarily the best long term solution) [Discussion issue for future](https://github.com/nulib/repodev_planning_and_docs/issues/2469)
- Update preservation check with `md5` column
- __NOTE: for now - 1st iteration - the step function itself has been manually installed__

- [ ] Patch
- [ ] Minor
- [x] Major (backwards incompatible GraphQL API change)

# Steps to Test

1. Set up the environment
    ```
    devstack down
    devstack update
    devstack up -d meadow
    mix meadow.buckets.create
    ```
2. Start your dev environment server
3. Run an ingest sheet ingest
4. Verify that checksum hash is present in the GraphQL API
5. Verify that the copy checksum functionality on the preservation tab in the UI 
6. Perform a UI file uplopad
7. Verify that the copy checksum functionality on the preservation tab in the UI 
7. Verify that checksum hash is present in the GraphQL API
3. Launch the Minio console in a web browser
3. Upload an object to the `dev-ingest` (or `dev-uploads`) bucket
4. Click on the newly uploaded object to see its details
5. Make sure there are `computed-[hash]` and `computed-[hash]-last-modified` tags for each hash type (`md5` and `sha1`). If the uploaded object it large, it might take several seconds for the tags to show up.
6. Make sure the auto-tagging does _not_ happen on other buckets.
7. You can `devstack logs -f minio-checksum` if you want to see the log output from the hash-tagging webhook
8. Let the preservation check run, download it and observe the md5 column

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [x] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [x] Terraform changes
  - [x] Adds/requires new or changed Terraform variables [`fixity_function`](https://github.com/nulib/tfvars/commit/9a6ece5f8d6dabc7f7063f775d58dd38c5c0d021) - we still need prod added to `tfvars`
- [x] Pipeline configuration changes (requires `mix meadow.pipeline.setup` run)
- [x] Other specific instructions/tasks
  -  **__Do not merge until after both #2682 and nulib/devstack#9 have been merged__**
  - requires all developers to run `devstack update` and restart the stack before using meadow again
  - Manually install and configure [Serverless Fixity for Digital Preservation Compliance](https://docs.aws.amazon.com/solutions/latest/serverless-fixity-for-digital-preservation-compliance/welcome.html) on production

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/lib/meadow/pipeline.ex
+++ b/lib/meadow/pipeline.ex
@@ -4,11 +4,22 @@ defmodule Meadow.Pipeline do
   """
   use Sequins.Pipeline
 
+  alias Meadow.Config
   alias Meadow.Data.{ActionStates, FileSets}
   alias Meadow.Data.Schemas.FileSet
   alias Meadow.Pipeline.Actions.Dispatcher
+  alias Meadow.Utils.AWS
+
+  import WaitForIt
 
   def ingest_file_set(attrs \\ %{}) do
+    case wait_for_checksum_tags(attrs) do
+      :ok -> do_ingest_file_set(attrs)
+      {:error, changeset} -> {:error, changeset}
+    end
+  end
+
+  defp do_ingest_file_set(attrs) do
     case FileSets.create_file_set(attrs) do
       {:ok, file_set} ->
         kickoff(file_set, %{role: file_set.role.id})
@@ -29,6 +40,26 @@ defmodule Meadow.Pipeline do
 
     with initial_action <- List.first(actions) do
       initial_action.send_message(%{file_set_id: file_set_id}, context)
+    end
+  end
+
+  defp wait_for_checksum_tags(%{core_metadata: %{location: location}} = attrs) do
+    with %{host: bucket, path: "/" <> key} <- URI.parse(location) do
+      case wait(AWS.check_object_tags!(bucket, key, Config.required_checksum_tags()),
+             timeout: Config.checksum_wait_timeout(),
+             frequency: 1_000
+           ) do
+        {:ok, true} ->
+          :ok
+
+        {:timeout, timeout} ->
+          {:error,
+           FileSet.changeset(attrs)
+           |> Ecto.Changeset.add_error(
+             :checksums,
+             "Timed out after #{timeout}ms waiting for checksum tags"
+           )}
+      end
     end
   end
 end
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -143,8 +143,16 @@ jobs:
           MINIO_ROOT_PASSWORD: minio123
           MINIO_ACCESS_KEY: minio
           MINIO_SECRET_KEY: minio123
+          MINIO_NOTIFY_WEBHOOK_ENABLE_checksum: on
+          MINIO_NOTIFY_WEBHOOK_ENDPOINT_checksum: http://minio-checksum:8000/fixity
         ports:
           - 9002:9000
+      minio-checksum:
+        image: nulib/minio-checksum
+        env:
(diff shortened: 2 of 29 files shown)


kdid:
This works great. Not approving yet since merge is on hold.

lib/meadow/pipeline.ex
```diff
@@ -31,4 +42,24 @@ defmodule Meadow.Pipeline do
       initial_action.send_message(%{file_set_id: file_set_id}, context)
     end
   end
+
+  defp wait_for_checksum_tags(%{core_metadata: %{location: location}} = attrs) do
+    with %{host: bucket, path: "/" <> key} <- URI.parse(location) do
+      case wait(AWS.check_object_tags!(bucket, key, Config.required_checksum_tags()),
```
bmquinn, line 48: Looking over the `wait_for_it` docs it looks like there's a convenience method [case_wait/3](https://hexdocs.pm/wait_for_it/WaitForIt.html#case_wait/3) that we might want to consider using here. I was thinking the bit:

```
The difference with case_wait is that if none of the clauses initially matches it will wait and periodically re-evaluate the clauses until one of them does match or a timeout occurs.
``` Happy to let it go but since we're using the library anyways it might be worthwhile. Thoughts?

kdid: We had `case_wait` in there but got frustrated with it for some reason, which I can't remember now. If you don't feel strongly I vote to leave as is.

mbklein: The difference between `case wait` and `case_wait` is that with the former, you can catch `{:timeout, timeout}` inside the same block as your `{:ok, return_value}` success clauses, while with `case_wait` you put your bare `return_value` clauses inside the `case_wait` and catch the `{:timeout, timeout}` outside of it. So you'd have to wrap `case_wait` in another `case` or an `if` block, which I think is super ugly.

bmquinn: Works for me!

bmquinn (approved):

PR #2949, opened 2022-02-28
Invalidate poster cache after poster generation

kdid:
# Summary 

Fixes a bug where resetting a poster image does not update in the UI because the image has been cached. 

# Specific Changes in this PR

- Invalidate the Cloudfront cache for the poster key after poster image is generated
- Should be skipped in the dev/test environment
- Adds variables for `iiif_cloudfront_distribution_id` 
- Gives Meadow task permissions to create cloudfront invalidation

**Terraform changes applied on staging

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

- Pull changes from `nulib/tfvars`

Dev: 
- We don't cache in the dev environment so hard to test
- But you should see in the logs that it's skipping invalidation if you generate a new poster image

Staging/Prod:
- set a new poster image on a file set that already has a poster image
- wait a minute or so and refresh the page
- expect to see the new poster image reflected in the UI

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [x] Terraform changes **Terraform changes applied on staging
  - [x] Adds/requires new or changed Terraform variables (tfvars)

# Tested/Verified
- [ ] End users/stakeholders


config/releases.exs
```diff
@@ -99,6 +99,8 @@ config :meadow,
   },
   environment: :prod,
   digital_collections_url: get_required_var.("DIGITAL_COLLECTIONS_URL"),
+  iiif_cloudfront_distribution_id: get_required_var.("IIIF_CLOUDFRONT_DISTRIBUTION_ID"),
+  iiif_cloudfront_version: get_required_var.("IIIF_CLOUDFRONT_VERSION"),
```
mbklein, line 103: I think a module attribute would be better for this than a variable, if you don't want to hardcode it. These AWS API version strings don't change very often at all, and there's no penalty for using an old one as long as your data conforms to the shape of that version. Which it will unless we change it, at which point we would presumably change the hardcoded version string as well.

As another example, we have `"Version": "2012-10-17"` hardcoded in our IAM policy documents all over our Terraform code.

mbklein (changes requested):

config/releases.exs
kdid: @mbklein - Updated, PTAL. Terraform reapplied on staging

mbklein: `releases.exs:103` still needs to go away or the container won't start due to the missing `IIIF_CLOUDFRONT_VERSION` variable.

mbklein (changes requested):

kdid:
@mbklein - oops forgot to save. Pushed.

mbklein (approved):

PR #2950, opened 2022-02-28
Upload file sets using their native content type as determined by the browser

mbklein:
# Summary 
Fix zero-byte browser uploads to Localstack

# Specific Changes in this PR
- Use a `Content-Type` of `currentFile.type` instead of `"multipart/form-data"` when uploading FileSet through the UI
- Remove double-dot from presigned file set URLs

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

Attach an audio, video, or image preservation file to a work using the UI and make sure it passes the `ExtractMimeType` action.

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/assets/js/components/Work/Tabs/Preservation/FileSetModal.jsx
+++ b/assets/js/components/Work/Tabs/Preservation/FileSetModal.jsx
@@ -172,7 +172,7 @@ function WorkTabsPreservationFileSetModal({
     });
 
     request.open("put", presignedUrl);
-    request.setRequestHeader("Content-Type", "multipart/form-data");
+    request.setRequestHeader("Content-Type", "application/octet-stream");
     request.send(currentFile);
     setStateXhr(request);
   };
--- a/lib/meadow/utils/aws.ex
+++ b/lib/meadow/utils/aws.ex
@@ -38,7 +38,7 @@ defmodule Meadow.Utils.AWS do
   end
 
   def presigned_url(bucket, %{upload_type: "file_set", filename: filename}) do
-    generate_presigned_url(bucket, "file_sets/#{Ecto.UUID.generate()}.#{Path.extname(filename)}")
+    generate_presigned_url(bucket, "file_sets/#{Ecto.UUID.generate()}#{Path.extname(filename)}")
   end
 
   def presigned_url(bucket, %{upload_type: "ingest_sheet"}) do


kdid:
@mbklein - Some file types seem to be working now but I'm still getting the error with .mkv files.

I didn't check the whole list but we need to be able to support these: https://github.com/nulib/repodev_planning_and_docs/issues/1924

I think I recall there was a reason we didn't go with the file type the browser thought it was in this case. (Because it didn't recognize all that we need to support).

kdid:
Note - Slightly different error: 

```
action=ExtractMimeType [info]  Extracting mime type
action=ExtractMimeType [error] Error extracting mime-type
action=ExtractMimeType [error] TypeError: buf.readUInt8 is not a function
action=ExtractMimeType [error] error in mime-type extraction
```

mbklein:
I've reproduced the error you're getting, but the cause of it seems to be completely different from the one we were getting yesterday. The file uploaded successfully this time, so the error is coming from farther down inside the MIME extraction code. But getting back to browser vs. S3 vs. localstack vs. mime lambda supported/known types, the important thing for _uploading_ seems to be just making sure localstack knows to receive and treat it as a binary file and not `multipart/form-data`. So the best option might be to hardcode it as `application/octet-stream` and move on. So I'll make that change.

I'm keeping this open while I investigate the new error.

mbklein:
This time, the upload isn't the problem. This error is coming from the `file-type` library. I can even reproduce it by taking S3/localstack out of the equation and telling `file-type` to use its own file reader:

```
➜ node --experimental-repl-await
Welcome to Node.js v14.19.0.
Type ".help" for more information.
> const FileType = require("file-type");
undefined
> await FileType.fromFile('small.m4v');
{ ext: 'm4v', mime: 'video/x-m4v' }
> await FileType.fromFile('sample.mkv');
Uncaught TypeError: buf.readUInt8 is not a function
    at Object.get (meadow/priv/nodejs/mime-type/node_modules/token-types/lib/index.js:12:20)
    at FileTokenizer.peekNumber (meadow/priv/nodejs/mime-type/node_modules/strtok3/lib/AbstractTokenizer.js:63:22)
    at async readField (meadow/priv/nodejs/mime-type/node_modules/file-type/core.js:680:16)
    at async readElement (meadow/priv/nodejs/mime-type/node_modules/file-type/core.js:695:15)
    at async _fromTokenizer (meadow/priv/nodejs/mime-type/node_modules/file-type/core.js:717:14)
    at async Object.fromFile (meadow/priv/nodejs/mime-type/node_modules/file-type/index.js:8:10)
    at async REPL2:1:24
```

This error occurs on every `.mkv` file I've tried.

I'm not sure what to do here except maybe rethink how we handle files the `file-type` package can't process. I don't think we can trust the extension without any further validation, because that defeats the purpose of trying to type it based on content in the first place.

kdid:
@mbklein - I just tried an .mkv on staging and mime-type extraction failed. So is this a known issue? 

Do all the other file types we need to support work in dev? If so, maybe we just open an issue about the .mkv's...

kdid (approved):

PR #2975, opened 2022-03-17
Adjust regex to ensure a starts with on URL pattern protocol.

mathewjordan:
# Summary 
URL Pattern matching was checking for any part of the string input into the URL field to start with `http://` or `https://`. In that case an input value of `foo https://bar.baz` would validate TRUE. This now validates false and will require a url to match a pattern `https://foo.bar` with only valid starts:

- `http://foo.bar`
- `https://foo.bar`
- `www.foo.bar`

_My solution here probably could be done completely with RegEx but I'm not sure how considering we are seeking to allow **www** as well as the protocols to be valid._

# Specific Changes in this PR
- Adjusts URL validation to make sure URL strings start with either `http://`, `https://`, or `www`

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

1. Edit any work
2. Add a Related URL of `foo https://bar.baz`
3. This should fail validation ❌ 
2. Add a Related URL of `https://foo.bar`
3. This should pass validation ✅ 

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/assets/js/services/helpers.js
+++ b/assets/js/services/helpers.js
@@ -1,6 +1,6 @@
 import moment from "moment";
 import { toast } from "bulma-toast";
-import { URL_PATTERN_MATCH } from "./global-vars";
+import { URL_PATTERN_MATCH, URL_PATTERN_START } from "./global-vars";
 import edtf from "edtf";
 
 /**
@@ -24,7 +24,10 @@ export function formatSimpleISODate(date) {
 }
 
 export function isUrlValid(url) {
-  return url.match(URL_PATTERN_MATCH) ? true : false;
+  return url.match(URL_PATTERN_MATCH) &&
+    URL_PATTERN_START.some((validStart) => url.startsWith(validStart))
+    ? true
+    : false;
 }
 
 export function isEDTFValid(edtfString) {
--- a/assets/js/services/helpers.js
+++ b/assets/js/services/helpers.js
@@ -24,10 +24,10 @@ export function formatSimpleISODate(date) {
 }
 
 export function isUrlValid(url) {
-  return url.match(URL_PATTERN_MATCH) &&
+  return (
+    url.match(URL_PATTERN_MATCH) &&
     URL_PATTERN_START.some((validStart) => url.startsWith(validStart))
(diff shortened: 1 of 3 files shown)


mathewjordan:
@adamjarling For your eyes when you return.

adamjarling (changes requested):
@mathewjordan This looks good for Batch Edit, however can we apply the same error message display to a single Work Edit?   

To test: in Work Edit, enter "asdf http://google.com" for "Related Url".  The form doesn't submit (which is probably good), but guessing the validation error just isn't getting communicated in the UI?

assets/js/services/helpers.js
```diff
@@ -24,7 +24,10 @@ export function formatSimpleISODate(date) {
 }
 
 export function isUrlValid(url) {
-  return url.match(URL_PATTERN_MATCH) ? true : false;
+  return url.match(URL_PATTERN_MATCH) &&
+    URL_PATTERN_START.some((validStart) => url.startsWith(validStart))
+    ? true
```
adamjarling, line 29: You probably don't even need lines 29-30 here.  Guessing it'll coerce to truthy?

mathewjordan (commented):
@adamjarling I dug into this a bit more to try and remember what we are doing here. It looks like all the validation here happens during the attempted submission process. Do we want to switch this up so that some minimal feedback is given during each onChange?

adamjarling:
@mathewjordan Either way as long as it's good UX, I'm fine pre-handling the submission or the alternative was React Hook Form catches the validation error and displays the error message prior to sending off the network request?

PR #3113, opened 2022-06-03
Configuration Cleanup

mbklein:
# Summary 

While trying to figure out a way to clean up all the `prefix` constants at the top of every config file, I realized that there were still a few references to Sequins in there, so I created a branch to clean them up. While I was at it, I created a helper module to replace a lot of the repetitive and convoluted Hush code used to configure the app dynamically.

This led to a chicken-and-egg problem, in which the app can't be compiled without the config, but the config depends on a module in the app. The only way around it I could find was to create _two_ modules – `Env.CompileTime` and `Env.Runtime` to handle the same functions in Mix mode and release mode, respectively. The `CompileTime` module is just a dynamically require-able version of the `RunTime` module. The only reason they have different names is to prevent a compile-time warning about redefining the module.

Honestly, the `Env`-module juggling doesn't really thrill me, and I'd love to find a better solution. But I think it's more than made up for by how much cleaner the config files are.

# Specific Changes in this PR
- Remove remaining Sequins references
- Move pipeline action configurations from top-level `:meadow` to `:meadow, Meadow.Pipeline`
- Create `lib/env.ex` (runtime config helper) and `config/env.exs` (compile-time config helper) to simplify config files
- Replace dynamic configuration tuples and prefix interpolations in config files with calls to the config helper module

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

- Run meadow. Make sure everything compiles and works as expected. Nothing specific beyond that.
- The release has been tested to the greatest extent possible without actually pushing a container image to staging

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders


app/config/config.exs
```diff
@@ -223,14 +214,14 @@ config :authoritex,
     Authoritex.Getty.TGN,
     Authoritex.Getty.ULAN,
     Authoritex.Getty,
-    Authoritex.LOC.Languages,
-    Authoritex.LOC.Names,
-    Authoritex.LOC.SubjectHeadings,
+    Authoritex.LOLanguages,
```
bmquinn, line 217: Looks like there's an accidental edit on these authority modules.

bmquinn (approved):

PR #3162, opened 2022-08-17
Add LastModified property to IIIF manifests

mbklein:
# Summary 
Add `LastModified` metadata property to IIIF manifests

# Specific Changes in this PR

- Add property function to IIIF modules
- Add metadata to IIIF v2 and v3 manifests

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

Generate a manifest and check the top-level `metadata` section

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/app/lib/meadow/data/schemas/iiif/V2/work.ex
+++ b/app/lib/meadow/data/schemas/iiif/V2/work.ex
@@ -7,6 +7,9 @@ defimpl Meadow.IIIF.V2.Resource, for: Meadow.Data.Schemas.Work do
       id: IIIF.V2.manifest_id(work.id),
       label: work.descriptive_metadata.title,
       description: work.descriptive_metadata.description,
+      metadata: [
+        IIIF.V2.property("LastModified", DateTime.utc_now() |> DateTime.to_iso8601())
+      ],
       sequences: [
         %Sequence{
           canvases:
--- a/app/lib/meadow/iiif/v2/generator.ex
+++ b/app/lib/meadow/iiif/v2/generator.ex
@@ -28,9 +28,11 @@ defmodule Meadow.IIIF.V2.Generator do
     |> Jason.encode!(pretty: true)
   end
 
+  defp to_json(manifest) when is_struct(manifest),
+    do: manifest |> Map.from_struct() |> to_json()
+
   defp to_json(manifest) do
     manifest
-    |> Map.from_struct()
     |> Enum.map(&set_property/1)
     |> Enum.filter(&filter_property/1)
     |> Enum.into(%{})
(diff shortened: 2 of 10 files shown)


app/lib/meadow/data/schemas/iiif/V2/work.ex
```diff
@@ -7,6 +7,9 @@ defimpl Meadow.IIIF.V2.Resource, for: Meadow.Data.Schemas.Work do
       id: IIIF.V2.manifest_id(work.id),
       label: work.descriptive_metadata.title,
       description: work.descriptive_metadata.description,
+      metadata: [
+        IIIF.V3.property("LastModified", DateTime.utc_now() |> DateTime.to_iso8601())
```
kdid, line 11: Did you mean to use the V2 `property` function here? It looks like you made two separate ones.

app/lib/meadow/iiif/v2/generator.ex
```diff
@@ -28,9 +28,11 @@ defmodule Meadow.IIIF.V2.Generator do
     |> Jason.encode!(pretty: true)
   end
 
+  defp to_json(manifest) when is_struct(manifest),
+    do: manifest |> Map.from_struct() |> to_json()
+
   defp to_json(manifest) do
     manifest
-    |> Map.from_struct()
```
kdid, line 33: Why are we removing this?

app/lib/meadow/iiif/v2/generator.ex
```diff
@@ -28,9 +28,11 @@ defmodule Meadow.IIIF.V2.Generator do
...
     |> Enum.map(&set_property/1)
     |> Enum.filter(&filter_property/1)
     |> Enum.into(%{})
```
mbklein, line 38: After I added the metadata property map, the generator started failing because it could _only_ handle structs, not bare maps. So I moved the `Map.from_struct()` bit to a second function head. The other solution would be to create a full struct definition for properties with `label` and `value` members, but even then, `to_json/1` would still need to handle bare maps because of the language keys in both the `label` and the `value`.

kdid (approved):

PR #3246, opened 2023-04-18
Reconfigure index constants

kdid:
- Expose constants for all 3 indices - default to work index if none provided.


adamjarling (commented):
@kdid Do we need to update the `miscellany` repo, or anything else to make this work?

adamjarling (approved):

kdid:
No, because what I realized when I actually looked at the code is that these aren't actual environment variables in that sense, they are defined right in the webpack config file. I notice that you currently do have at least one query that would need an index besides the work index (homepage - collection count something-or-rather) so I didn't just want to straight up swap in just the one index. Let me know if this makes sense.

PR #3312, opened 2023-04-27
Adds more granular "reindex all" functionality

bmquinn:
# Summary 
You can now pass in any combination of versions and schemas to the `reindex_all/1` function. See test for examples.

# Specific Changes in this PR
- indexer and indexer_test updates

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test
Please let end users know what they need to do to test on staging or production

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [x] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/app/lib/meadow/data/indexer.ex
+++ b/app/lib/meadow/data/indexer.ex
@@ -32,11 +32,18 @@ defmodule Meadow.Data.Indexer do
   end
 
   def reindex_all(version) do
-    [FileSet, Work, Collection]
-    |> Enum.each(fn schema ->
-      SearchClient.hot_swap(schema, version, fn index ->
-        synchronize_schema(schema, version, index, :all)
-      end)
+    reindex_all(version, [FileSet, Work, Collection])
+  end
+
+  def reindex_all(version, schemas) when is_list(schemas) do
+    schemas
+    |> Enum.uniq()
+    |> Enum.each(&reindex_all(version, &1))
+  end
+
+  def reindex_all(version, schema) do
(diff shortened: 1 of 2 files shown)


app/lib/meadow/data/indexer.ex
```diff
@@ -40,6 +40,18 @@ defmodule Meadow.Data.Indexer do
     end)
   end
```
mbklein, line 42: Is it maybe worth rewriting `reindex_all/1` to call your new `reindex_all/2` instead of having the code repeated?

```suggestion
  def reindex_all(version) do
    reindex_all(2, [FileSet, Work, Collection])
  end
```

You'd need to leave the special case `reindex_all(1)` alone since the hot swap is outside the loop, but that's going away soon anyway.

mbklein (approved):
Looks great! One small suggestion and then good to merge.

app/lib/meadow/data/indexer.ex
bmquinn: Thanks for calling this out, hopefully that does the trick.

mbklein (approved):

PR #3325, opened 2023-05-04
Switch from webpack to esbuild

mbklein:
# Summary 

Switch Meadow front end from webpack to esbuild

# Specific Changes in this PR
- Configure esbuild for `npm run deploy` and `npm run watch`
- Fix issue in `elasticsearch.js` that was preventing it from building under esbuild
- Configure Meadow to use esbuild as its file watcher

- list changes

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

- Make sure the entire front end works in dev
- We won't be able to test the packaged build until it's on staging

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders


app/assets/build.js
```diff
@@ -0,0 +1,95 @@
...
+  outdir: "../priv/static/js",
+  logLevel: "info",
+  sourcemap: "linked",
+  define,
+  loader,
+  plugins,
+};
+
+async function watchMode(opts) {
+  const ctx = await esbuild.context({ ...opts, sourcemap: "inline" });
```
adamjarling, line 66: This will be great if we can get sourcemaps in `dev` watch mode.

app/assets/js/services/elasticsearch.js
```diff
@@ -7,9 +7,25 @@ export const ELASTICSEARCH_PROXY_ENDPOINT = `${host}:${port}/_search`;
 export const ELASTICSEARCH_INDEX_NAME = __ELASTICSEARCH_INDEX__;
 
 const Elasticsearch = require("elasticsearch");
+
```
adamjarling, line 10: I'd love to get `Elasticsearch` out of the client-side after we push this through, as part of the "Meadow re-evaluate" batch of client-side app work.   As long as this works and queries are firing, even if just limping along or not pretty, we'll extract it soon enough and don't need to spend too much headspace on it.

adamjarling (approved):
Let it rip 🚀. Nice work @mbklein

app/assets/js/services/elasticsearch.js
mbklein: I agree. The newer version of the official JS client even says that [in-browser use is not supported](https://github.com/elastic/elasticsearch-js#browser).

PR #3355, opened 2023-05-19
Fix index version issue, thumbnails and batch works link

kdid:
Fixes #2842 and #2843

# Summary 

- Fix thumbnails on search page. (They had doubled up dimensions, ex: `/square/500,500/0/default.jpg/square/500,500/0/default.jpg`)
- Fix link from batch edit dashboard list to view works affected by a batch
- Remove `@v2_work_index` attributes and get value directly since it is incorrect at build time
- Remove v1 index configuration from `releases.exs`

### DIFF
--- a/app/assets/js/components/Work/CardItem.jsx
+++ b/app/assets/js/components/Work/CardItem.jsx
@@ -7,7 +7,6 @@ import React from "react";
 import { Tag } from "@nulib/design-system";
 import UIVisibilityTag from "@js/components/UI/VisibilityTag";
 import UIWorkImage from "@js/components/UI/WorkImage";
-import { getIIIFImageUrl } from "@js/services/helpers";
 
 const breakWord = css`
   word-break: break-all;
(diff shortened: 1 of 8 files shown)


adamjarling (approved):
Looks good.

app/assets/js/components/Work/CardItem.jsx
```diff
@@ -29,7 +29,7 @@ const WorkCardItem = ({
         <figure className="image is-3by3">
           <Link to={`/work/${id}`}>
             <UIWorkImage
-              imageUrl={getIIIFImageUrl(representativeImage)}
```
mbklein, line 32: It looks like this PR is removing the only two references to this function, so it might make sense to remove the imports in both files and also the function itself from `js/services/helpers.js`

mbklein (approved):
Approving with one optional suggestion

app/assets/js/components/Work/CardItem.jsx
kdid: Done

PR #3422, opened 2023-06-28
Add track_total_hits flag to all msearch requests

kdid:
# Summary 

Works around sorting bug in UI/Reactivesearch https://github.com/nulib/repodev_planning_and_docs/issues/4007

# Specific Changes in this PR
- Adds `track_total_hits=true` to all `_msearch` requests if not present.

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

- Make sure all searches in Meadow UI work and result count is preserved even after sorting results

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/app/lib/meadow_web/plugs/search.ex
+++ b/app/lib/meadow_web/plugs/search.ex
@@ -41,6 +41,7 @@ defmodule MeadowWeb.Plugs.Search do
     case HTTP.request(method, path <> query_string, body, []) do
       {:ok, response} ->
         conn
+        |> assign(:updated_req_body, body)
         |> put_resp_content_type("application/json")
         |> resp(200, Jason.encode!(response.body))
         |> send_resp()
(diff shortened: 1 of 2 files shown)


app/lib/meadow_web/plugs/search.ex
```diff
@@ -68,10 +69,36 @@ defmodule MeadowWeb.Plugs.Search do
...
+      body
+      |> String.trim()
+      |> String.split("\n")
+      |> Enum.map_join("\n", fn line ->
+        line
+        |> Jason.decode!()
+        |> add_track_total_hits()
+        |> Jason.encode!()
+      end)
+      |> Kernel.<>("\n")
```
mbklein, line 90: Oh cool. I didn't even know this was a thing.

app/lib/meadow_web/plugs/search.ex
```diff
@@ -68,10 +69,36 @@ defmodule MeadowWeb.Plugs.Search do
...
+    else
+      body
```
mbklein, line 92: Don't we want to add `track_total_hits` to a single `_search` / `application/json` query as well? I can't remember if there's a reason we're limiting this to `_msearch`. (I get that we're going with what ReactiveSearch is actually doing, but I'm just asking for the sake of consistency.)

Seems like it would be easy enough to add
```suggestion
      body
      |> Jason.decode!()
      |> add_track_total_hits()
      |> Jason.encode!()
```
right here.

kdid: Added

mbklein (approved):

PR #3766, opened 2024-01-10
Refetch token for the work provider using the `expires` datetime.

mathewjordan:
# Summary 
On the Work component, this removes the polling on that was previously used in the Apollo Client useQuery() for **GET_DC_API_TOKEN**. The token is now [refetched](https://www.apollographql.com/docs/react/data/refetching) when the token is 1 minute (60000ms) from expiring.

This work also revises a re-rendering issue caused by UI/IIIF/Viewer component infinitely dispatching the activeFileSet from the Clover canvasIdCallback. This now only dispatches if the activeFileSet ID does not match.

# Specific Changes in this PR
- Revises useQuery() function to use `refetch` function instead of polling interval
- Fixes re-rerender bug caused by unchecked conditional in IIIF Viewer component

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test
- Your meadow environment should have a video or audio work with a time of longer than 5 minutes (the current expire time for the token).
- Go to the Work, begin playing the video or audio work.
- Wait 5 minutes, you can `console.log(currentDate, expiresDate, expireDifference, token);` in the  
`handleTokenUpdate()` to watch things...
- The video/audio file should continue playing after the token updates (and to the end)

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/app/assets/js/components/Work/Work.tsx
+++ b/app/assets/js/components/Work/Work.tsx
@@ -15,21 +15,35 @@ const Work = ({ work }: { work: WorkType }) => {
   /**
    * Get the DC API super user token from the API every 5 minutes.
    */
-  const { data: dataDcApiToken, error: errorDcApiToken } = useQuery(
-    GET_DC_API_TOKEN,
-    { pollInterval: 300000 },
-  );
+  const {
+    data: dataDcApiToken,
+    error: errorDcApiToken,
+    refetch: refetchDcApiToken,
+  } = useQuery(GET_DC_API_TOKEN);
 
-  const dcApiToken = dataDcApiToken?.dcApiToken?.token;
+  const { token, expires } = dataDcApiToken?.dcApiToken || {};
 
-  useEffect(
-    () =>
+  useEffect(() => {
+    if (token) {
       workDispatch({
         type: "updateDcApiToken",
-        dcApiToken,
-      }),
-    [dcApiToken],
-  );
+        dcApiToken: token,
+      });
+    }
+
+    const handleTokenUpdate = setInterval(() => {
+      const timePadding = 60000; // 1 minute
+      const currentDate = new Date();
+      const expiresDate = new Date(expires);
+      const expireDifference = expiresDate.getTime() - currentDate.getTime();
+      
+      if (expireDifference < timePadding) {
+        refetchDcApiToken();
+      }
+    }, 500);
+
+    return () => clearInterval(handleTokenUpdate);
+  }, [token]);
 
   if (errorDcApiToken) console.error(errorDcApiToken);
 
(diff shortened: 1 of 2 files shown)


app/assets/js/components/Work/Work.tsx
```diff
@@ -15,21 +15,35 @@ const Work = ({ work }: { work: WorkType }) => {
...
+    const handleTokenUpdate = setInterval(() => {
+      const timePadding = 60000; // 1 minute
+      const currentDate = new Date();
+      const expiresDate = new Date(expires);
+      const expireDifference = expiresDate.getTime() - currentDate.getTime();
+      console.log(currentDate, expiresDate, expireDifference);
+      if (expireDifference < timePadding) {
+        refetchDcApiToken();
+      }
+    }, 100);
```
adamjarling, line 43: @mathewjordan Does this need to be such a short interval?  Looks like the component re-renders are reasonable 👍, but this interval is firing continuously...not sure how much of a performance hit, if any it is tho...

mathewjordan: We need some sort of interval here. It could be a little looser, but all that is happening is the conditional on line 40 which shouldn't be too intensive.

adamjarling: Yea, resource intensive probably not that big a hit but it is something just running non-stop which fires up my OCD 😀.  What if we either bumped that interval value a touch (`300`?), _or_ what about something more targeted like this?

```
useEffect(() => {
    console.log("useEffect()");

    let handleTokenUpdate: NodeJS.Timeout | null = null;

    if (token) {
      workDispatch({
        type: "updateDcApiToken",
        dcApiToken: token,
      });

      const timePadding = 60000; // 1 minute
      const currentDate = new Date();
      const expiresDate = new Date(expires);
      const expireDifference = expiresDate.getTime() - currentDate.getTime();

      if (expireDifference > timePadding) {
        console.log(
          "expireDifference > timePadding",
          expireDifference,
          timePadding,
        );
        handleTokenUpdate = setTimeout(() => {
          console.log("refetchDcApiToken()", expireDifference - timePadding);
          refetchDcApiToken();
        }, expireDifference - timePadding);
      }
    }

    return () => {
      if (handleTokenUpdate) {
        clearTimeout(handleTokenUpdate);
      }
    };
  }, [token, expires, refetchDcApiToken, workDispatch]);
```

adamjarling: Convenience `console.logs` left in above...

mathewjordan:
@adamjarling Made your suggestion and upped the interval. Did a half-second (500ms).

adamjarling (approved):

PR #3774, opened 2024-01-17
Use CSV manifests instead of random IDs for export

mbklein:
# Summary 

Use CSV manifests instead of random IDs for export

# Specific Changes in this PR
- Change `Meadow.Seed.Export.export_ingest_sheets/3` and `Meadow.Seed.Export.export_standalone_works/3` to take a list of IDs instead of a number of IDs to pick at random
- Change `Mix.Tasks.Meadow.Seed.Export` to take CSV files with ingest sheet and work IDs as parameters instead of integers

# Version bump required by the PR

See [Semantic Versioning 2.0.0](https://semver.org/) for help discerning which is required.

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

- Create a CSV with a couple ingest sheet IDs in the first column and another with a few work IDs in the first column
- Run `mix meadow.seed.export --ingest-sheets ingest_sheet_export.csv --works works_export.csv --bucket SOME_BUCKET --prefix SOME/PATH/`
- When the process is finished, run `aws s3 ls --recursive s3://SOME_BUCKET/SOME/PATH/` to make sure there's export data there

Try again but omit the `--ingest-sheets` or `--works` parameter to make sure it skips that step (by exporting 0 of the given type of asset).

If you want to be truly ambitious, you can wipe your data and empty your buckets and run `mix meadow.seed.import --prefix s3://SOME_BUCKET/SOME/PATH/` to re-import the data you just exported, but obviously you'll lose all the resources that _weren't_ exported. But since the only change to the export functionality is where the IDs come from, making sure the data gets exported properly is more important than testing the import.

# :rocket: Deployment Notes

**Note** - __if you check any of these boxes go to the [(always open) `main` <- `staging` PR](https://github.com/nulib/meadow/pulls) and add detailed notes and instructions to help out others who may end up deploying your changes to production__

- Backward compatible API changes
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] CSV metadata export/update API
  - [ ] Shared Links export API
- Backwards-incompatible API changes
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] CSV metadata export/update API
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires database triggers disabled during deployment/migration
- [ ] Requires reindex
- [ ] Terraform changes
[...]

PR #3800, opened 2024-01-30
Move export/import business logic out of tasks and into modules

mbklein:
# Summary 

This PR moves all of the seed export/import business logic (loading CSVs, resolving prefixes, checking conditions) out of the `Mix.Tasks.Meadow.Seed.{Export, Import}` modules and into the `Meadow.Seed.{Export, Import}` modules so that they can be run from Livebook. It also updates the inputs to the `export` process so that the CSVs containing the IDs of resources to export can be streamed directly from S3 instead of having to reside on the local disk, which also makes it easier to run the process from Livebook.

# Specific Changes in this PR
- Refactor `Meadow.Seed.Export.export/1` and `Meadow.Seed.Import.import/1` to take keyword lists
- Move the logic of applying defaults and interpreting those keyword lists from the tasks into the modules
- Keep the tasks as command line parsing wrappers for the modules

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test
Same as #3774, with a slight change:

- Create a CSV with a couple ingest sheet IDs in the first column and another with a few work IDs in the first column
- Run `mix meadow.seed.export --ingest-sheets ingest_sheet_export.csv --works works_export.csv --bucket SOME_BUCKET --prefix SOME/PATH/`
- When the process is finished, run `aws s3 ls --recursive s3://SOME_BUCKET/SOME/PATH/` to make sure there's export data there
- Repeat but omit the `--ingest-sheets` or `--works` parameter to make sure it skips that step (by exporting 0 of the given type of asset).
- Repeat but use `file:` URIs for the CSVs (e.g., `file:///path/to/ingest_sheet_export.csv`)
- Upload the CSVs to a bucket and use `s3://bucket/path/to/csv` URIs for the CSVs
- Try all of the above again, but this time do it from within `iex` to test the direct module call:
    ```
    Meadow.Seed.Export.export(ingest_sheets: "ingest_sheet_export.csv",
      works: "works_export.csv", bucket: "SOME_BUCKET", prefix: "SOME/PATH/`)
    ```

If you want to be truly ambitious, you can wipe your data and empty your buckets and run `mix meadow.seed.import --prefix s3://SOME_BUCKET/SOME/PATH/` (or `Meadow.Seed.Import.import(prefix: ...)`) to re-import the data you just exported, but obviously you'll lose all the resources that _weren't_ exported.

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/app/lib/meadow/seed/export.ex
+++ b/app/lib/meadow/seed/export.ex
@@ -238,18 +238,18 @@ defmodule Meadow.Seed.Export do
 
   defp missing?(value), do: is_nil(value) or value == ""
 
+  defp get_ids(nil), do: []
+  defp get_ids(""), do: []
   defp get_ids("s3://" <> _ = url), do: ids_from_csv(url)
   defp get_ids("file://" <> _ = url), do: ids_from_csv(url)
   defp get_ids("http://" <> _ = url), do: ids_from_csv(url)
   defp get_ids("https://" <> _ = url), do: ids_from_csv(url)
   defp get_ids(url), do: ids_from_csv("file://" <> url)
 
-  defp ids_from_csv(url) when is_binary(url) and byte_size(url) > 0 do
+  defp ids_from_csv(url) do
     Meadow.Utils.Stream.stream_from(url)
     |> CSV.parse_stream(skip_headers: false)
     |> Stream.map(fn [id | _] -> id end)
     |> Enum.to_list()
   end
-
-  defp ids_from_csv(_), do: []
 end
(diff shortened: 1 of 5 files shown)


bmquinn (approved):
This is a great pattern that we should keep in mind with new mix tasks going forward! 👍

PR #4391, opened 2025-02-05
PDF file sets should not have a pyramid location

kdid:
# Summary 

On the preservation report, Auxiliary files of type PDF were erroring because the pyramid was not found.

# Specific Changes in this PR

- Update `Meadow.Data.FileSets.pyramid_uri_for()` so that it does not return a pyramid location for a file set with `mime-type` of `application/pdf`

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

- Check preservation report for a pdf Auxiliary file. 
- The `pyramid_location` column should be empty and the `pyramid_exists` column value should be `N/A`

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/app/lib/meadow/data/file_sets.ex
+++ b/app/lib/meadow/data/file_sets.ex
@@ -291,21 +291,15 @@ defmodule Meadow.Data.FileSets do
   @doc """
   Get the pyramid path for a file set
   """
-  def pyramid_uri_for(%FileSet{role: %{id: "P"}}), do: nil
-  def pyramid_uri_for(%FileSet{role: %{id: "S"}}), do: nil
+  def pyramid_uri_for(%FileSet{role: %{id: "A"}, core_metadata: %{mime_type: "image/" <> _thing}} = file_set),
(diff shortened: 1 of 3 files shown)


mbklein (approved):
Approving this as the least invasive way to make this work, but I wonder if instead of excluding specific types from having pyramids, we should only include it for the specific types that need them? (e.g., `image/*` and anything with a zoomable thumbnail?)

kdid:
@mbklein - good catch. I made an adjustment. Thanks.

PR #4813, opened 2025-09-10
Switchover usage of UIFormSelect `disabled` to `aria-readonly`

mathewjordan:
# Summary 
This resolves an issue related to our form on the Add fileset modal losing its fileset role value prior submission. I determined that the `disabled` attribute on our Role `select` element was causing the form to deregister the field after a successful upload of a file to S3. To workaround this issue, we now opt for a custom isReadOnly prop which adjusts the select to act similar to a `disabled` attribute in making the element uneditable while retaining  its registration with the React Hook Form for the add fileset modal.

Note: The original issue seems very much an edge case and the recreation is highly dependent on the users' browser and the speed at which the upload to S3 occurs. I could only recreate it in Firefox while using the drag and drop feature with the Add Fileset `dropzone`. 

# Specific Changes in this PR
- Adds `isReadOnly` prop to the UIFormSelect component
- Replaces the `disabled` attribute of fileset role select with `isReadOnly`

- [ ] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

Best tested in Firefox.

- Go to any work. ( work type doesn't seem to matter)
- Go to preservation tab
- Without refreshing, Click add a fileset
  - Select any role (role does not seem matter...)
  - Pull in a file using the drag a drop
  - fill out accession and other fields
  - click ingest file
- Immediately without refreshing, Repeat step 3 above.
  - Role should be retained
  - Ingest should be successful

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/app/assets/js/components/UI/Form/Select.jsx
+++ b/app/assets/js/components/UI/Form/Select.jsx
@@ -2,47 +2,61 @@ import PropTypes from "prop-types";
 import React from "react";
 import { useFormContext } from "react-hook-form";
 
+/**
+ * `isReadOnly` simulates readonly for <select>:
+ * - Uses CSS to block interaction (keeps RHF registration + submission).
+ * - Avoid `disabled` if you need the value in RHF form state.
+ */
 const UIFormSelect = ({
   isReactHookForm,
   name,
   label,
-  // TODO: Clean up usages of UIFormSelect to use "hasErrors" instead of passing in "errors" object
   hasErrors,
   required,
   options = [],
   defaultValue,
   showHelper,
-  ...passedInProps
+  isReadOnly = false,
+  ...props
 }) => {
-  let errors = {},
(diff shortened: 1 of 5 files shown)


mbklein (approved):
Reproduced the issue in Firefox on `deploy/staging` branch, then switched to the PR branch and the issue did not occur. 👍🏻

PR #4886, opened 2025-10-15
Change terminology used for plans and plan changes

mbklein:
# Summary 
Change terminology used for plans and plan changes

# Specific Changes in this PR
- add **proposed** status to both `Plan` and `PlanChange` schema
- use **apply** instead of **execute**
- use **completed** instead of **executed**

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

Requires re-creating the `plans` and `plan_changes` tables in development:
```shell
mix do ecto.rollback, ecto.migrate
```

After merge, will require re-creating the same tables in the deployed Meadow AI instance, which can be done from Livebook.
```elixir
Meadow.Repo
|> Ecto.Migrator.with_repo(fn repo ->
  Ecto.Migrator.run(repo, :down, step: 1)
  Ecto.Migrator.run(repo, :up, all: true)
end)
```

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [x] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ Meadow supports AI agent-generated plans for batch modifications to works. The s
 - `query`: OpenSearch query string identifying target works
   - Collection query: `"collection.id:abc-123"`
   - Specific works: `"id:(work-id-1 OR work-id-2 OR work-id-3)"`
-- `status`: `:pending`, `:approved`, `:rejected`, `:executed`, or `:error`
+- `status`: `:pending, `:proposed`, `:approved`, `:rejected`, `:completed`, or `:error`
 
 **PlanChanges** - Work-specific modifications
 - `plan_id`: Foreign key to parent plan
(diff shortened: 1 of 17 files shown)


bmquinn:
The `proposed` status would only be at the `PlanChange` level, the `update_plan_change` tool Karen is working on will allow the agent to add the change along with setting the status from `pending` to `proposed`.

kdid:
This makes sense to me. I think rejecting a plan would set plan changes to rejected. Not sure either way on deleting pending changes. I can see why we don't want to keep them there in a pending state. But if we delete them I guess it makes me ask why not just delete all plan changes at that point (even rejected/approved ones)?

kdid:
Or maybe set all changes to rejected incl. approved/pending

mbklein:
I think this gets us far enough for me to finish this PR; we can refine what happens to specific changes as we implement the rest.

kdid (approved):

PR #4891, opened 2025-10-17
Add MeadowWeb.Schema.Public for use by the MeadowAI agent

bmquinn:
- Adds a new, scoped-down GraphQL schema for the agent to use

### DIFF
--- a/app/lib/meadow_web/router.ex
+++ b/app/lib/meadow_web/router.ex
@@ -70,6 +70,13 @@ defmodule MeadowWeb.Router do
       before_send: {Middleware.AssumeRole, :update_user_role}
     )
 
+    forward("/graphiql/agent", Absinthe.Plug.GraphiQL,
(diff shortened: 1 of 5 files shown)


app/lib/meadow_web/router.ex
```diff
@@ -70,6 +70,13 @@ defmodule MeadowWeb.Router do
       before_send: {Middleware.AssumeRole, :update_user_role}
     )
 
+    forward("/graphiqlai", Absinthe.Plug.GraphiQL,
```
mbklein: I can't really explain why, but I'd be happier with this namespaced.
```suggestion
    forward("/graphiql/ai", Absinthe.Plug.GraphiQL,
```

bmquinn:
> Question about this @bmquinn - do we want to still scope down the token so the agent would not be able to perform the other graphQL operations?

I think this is a separate issue, I'll write a ticket.

PR #5179, opened 2026-02-11
Convert metadata agent from PythonX to NodeJS Lambda

mbklein:
fixes https://github.com/nulib/repodev_planning_and_docs/issues/5755

# Summary 
Convert metadata agent from internal PythonX to deployed NodeJS Lambda

# Specific Changes in this PR
- Convert metadata agent from PythonX to NodeJS Lambda
- Use cast/send to make MetadataAgent handle multiple requests at once
- Move send_status_update and fetch_iiif_image into the Elixir MCP Server
- Catch errors in MCP Tools
- Port the metadata agent lambda to the TypeScript SDK for a smaller deploy footprint
- Add metadata lambda to pipeline infrastructure template
- Clean up lingering warnings

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test

- Start Meadow normally on the branch in the dev environment to test the deployed dev-environment version of the lambda
  - You should be able to load different works in two different windows and perform simultaneous Auto Edits
- Use `make pipeline-start` and start Meadow with `USE_SAM_LAMBDAS=true` to test the lambda locally
  - You can watch the SAM log to see the lambda think and use tools
  - SAM will only allow one execution per lambda at a time, so simultaneous edits won't work in local development mode

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Terraform changes
- [x] SAM lambda changes
  - [ ] Adds/requires new SAM variables

# Tested/Verified
- [ ] End users/stakeholders


bmquinn (approved):
Agent works in dev, tests passing 👍

kdid:
@mbklein - I tried this twice editing two different works in two different browsers at the same time. Each time one work was showing updates on screen and eventually produced a plan, but the other one stayed in "Initializing plan" and then eventually errored in the UI and this was seen in the logs

```
module=MeadowAI.MetadataAgent [error] Error executing Claude query: no function clause matching in MeadowAI.MetadataAgent.process_execution_result/1
module=MeadowWeb.Resolvers.Chat [error] AI query error: {:query_execution_error, %FunctionClauseError{module: MeadowAI.MetadataAgent, function: :process_execution_result, arity: 1, kind: nil, args: nil, clauses: nil}}
```

mbklein:
Odd. I'll check it out.

mbklein:
The concurrency issue was caused by the fact that `MeadowAI.MetadataAgent.query/2` was using `GenServer.call/3` (blocking) instead of `GenServer.cast/2` (non-blocking) to execute the lambda.

Before:
- `query/2` used `GenServer.call/3`, which waits for a return-value response
- `MetadataAgent.handle_call(:query...)` would execute the lambda, and block waiting for a response, allowing other messages to queue while it waited
- Additional calls to `query/2` would stack up, since `handle_call/3/` would only process 1 message at a time

Now:
- `query/2` uses `GenServer.cast/2` inside an ephemeral async process, then blocks until `opts[:timeout]` waiting to receive a response message
- `MetadataAgent.handle_cast(:query...)` executes the lambda inside a supervised async task, and returns immediately, clearing the mailbox and allowing additional messages to be processed immediately
- When the supervised task completes, it `send`s the result back to the caller
- Additional calls to `query/2` don't have to wait, since `handle_cast` keeps the mailbox empty

You'll see a `Task.await(:infinity)` inside `MetadataAgent.query/2`, because the task being waited on has its own `timeout` clause that limits how long `:infinity` can be.

bmquinn (approved):
Simultaneous auto-editing working for me in dev:

Image

kdid (approved):

PR #5237, opened 2026-02-27
Allow bulk file set additions to existing works on ingest

kdid:
# Summary 

fixes https://github.com/nulib/repodev_planning_and_docs/issues/5793

Users want to be able to attach file sets en masse to existing works. 

# Specific Changes in this PR

- Add optional `add_to_existing` (boolean) column to ingest sheet (backwards compatible)
- Adjust validation logic to error for add to existing file sets if target work is not found
- Adjust UI ingest sheet complete screen language and flow to handle file set additions

# Steps to Test

- Test ingest as in these fixtures (and/or other cases):
  - [Add to existing](https://github.com/nulib/meadow/blob/9fe04207126463bd5e85e6fcd512742c806e18b1/app/test/fixtures/ingest_sheet_add_to_existing.csv)
  - [Add to existing missing work](https://github.com/nulib/meadow/blob/9fe04207126463bd5e85e6fcd512742c806e18b1/app/test/fixtures/ingest_sheet_add_to_existing_missing_work.csv)
  - Existing ingests w/out the `add_to_existing` column should still be valid

- [x] Patch
- [ ] Minor
- [ ] Major

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [x] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Requires `terraform apply`
  - [ ] Adds/requires new or changed `ENVIRONMENT.tfvars` variables
- [ ] Requires `sam deploy`
  - [ ] Adds/requires new or changed `samconfig.yaml` variables

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/app/assets/js/components/IngestSheet/Completed.jsx
+++ b/app/assets/js/components/IngestSheet/Completed.jsx
@@ -12,7 +12,7 @@ import IngestSheetCompletedErrors from "./Completed/Errors";
 import UIPreviewItems from "../UI/PreviewItems";
 import { Button } from "@nulib/design-system";
 import { IconImages } from "@js/components/Icon";
-import { ActionHeadline, LevelItem, Skeleton } from "@js/components/UI/UI";
+import { ActionHeadline, Skeleton } from "@js/components/UI/UI";
 
 const IngestSheetCompleted = ({ sheetId, title }) => {
   const history = useHistory();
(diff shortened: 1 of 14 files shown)


bmquinn:
This looks great @kdid! The only question I had was really about workflow, not the code itself I think.

I'm trying to wrap my mind around what this change means for the ingest bucket organizational aspects like `Projects` and relative naming in the spreadsheet, etc. Is the idea that the file sets are always appended to works within an already existing Project as a unit? Or would the production staff be appending file sets to works that span across existing projects (or might not be associated with an ingest sheet b/c they were ingested via migration or manually)? Maybe new file sets will always be part of a previous ingest sheet, but I can imagine things getting confusing with `work_image` or `vtt` files being in different locations in the ingest bucket or something.

One other small thing I noticed is the preview page doesn't contain links to searches like ingest sheets usually do. Not sure if they'd want that here, too, but I noticed it makes it kind of harder to review the updates afterward:

SCR-20260227-iyzi

PR #5530, opened 2026-06-15
Mint new ARK for published/private work as reserved and then transition

### LINKED ISSUE #5529: ARK minting fails when a work is published and private
Every time a work is updated, Meadow checks to see (a) that it has an ARK, and (b) the ARK's published metadata (title, creator, publisher, publication year, resource type) is consistent with the updated work metadata. So (a) should take care of situations where the ARK service failed to generate an ARK on the first try by creating it on the fly as needed, but if the work has entered published/private state before that happens, it will never succeed.

The reason is that the EZID protocol requires new ARKs to be minted with a status of `reserved` or `public`. Unpublished works are always `reserved`, while published/public and published/institution are `public`, so those cases succeed. But published/private works try to assign a status of `unavailable | restricted`, which EZID only allows as a transition from an existing `reserved` or `public` state. The solution is to mint ARKs for published/private works as `reserved` and then immediately change them instead of trying to mint them as `unavailable | restricted` to begin with.

mbklein:
# Summary 
Mint new ARK for published/private work as reserved and then transition. Fixes #5529 

# Specific Changes in this PR
- Update `Meadow.Arks.mint_ark/1` with proper behavior
- Move `Meadow.Arks` module to correct directory

- [x] Patch
- [ ] Minor
- [ ] Major

# Steps to Test
Best done in IEx, as the UI doesn't provide the option of creating new works with `published: true`.

- Create a new work with `visibility: %{id: "RESTRICTED"}` and `published: true`
- Make sure new work has an ark
- Make sure `Meadow.Ark.get_from_source(work.descriptive_metadata.ark)` has `status: "unavailable | restricted"`
- Create new works with other values for visibility and published
  - ARK status should be `reserved` for unpublished works and `public` for non-restricted published works
- Change visibility and published status on existing works and make sure their new ARK statuses make sense

# :rocket: Deployment Notes

  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
  - [ ] Database Schema changes
  - [ ] GraphQL API
  - [ ] Elasticsearch API
  - [ ] Ingest Sheet
  - [ ] Shared Links export API
- [ ] Requires data migration
- [ ] Requires reindex
- [ ] Requires `terraform apply`
  - [ ] Adds/requires new or changed `ENVIRONMENT.tfvars` variables
- [ ] Requires `sam deploy`
  - [ ] Adds/requires new or changed `samconfig.yaml` variables

# Tested/Verified
- [ ] End users/stakeholders

### DIFF
--- a/app/lib/meadow/ark/mock_server.ex
+++ b/app/lib/meadow/ark/mock_server.ex
@@ -106,7 +106,7 @@ defmodule Meadow.Ark.MockServer do
       send_message({:post, :credentials, Plug.BasicAuth.parse_basic_auth(conn)})
       send_message({:post, :body, body})
 
-      case verify_metadata(body) do
+      case verify_metadata(ark, body) do
         :ok ->
           Cachex.put!(@cache, ark, body)
           send_resp(conn, 201, "success: #{ark}")
@@ -133,7 +133,7 @@ defmodule Meadow.Ark.MockServer do
         send_message({method, :body, body})
       end
 
-      case verify_metadata(body) do
+      case verify_metadata(ark, body) do
         :ok ->
           Cachex.put!(@cache, ark, body)
           send_resp(conn, 200, "success: #{ark}\n#{body}")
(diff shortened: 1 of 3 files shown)


bmquinn (approved):
Looks great 👍

kdid (approved):

PR #5578, opened 2026-07-02
Preserve file set order during work transfer

kdid:
# Summary
Users report that transferring a subset of file sets to another work (existing or newly created) scrambles their order. This fixes `Meadow.Data.Works.TransferFileSets` to preserve the file sets' actual saved order (rank) across the transfer, so a manually reordered Access/Auxiliary sequence survives moving to a new or existing work.

fixes https://github.com/nulib/repodev_planning_and_docs/issues/5829

# Specific Changes in this PR
- `transfer_fileset_subset/2` now re-derives file set order from the persisted `rank` field (grouped by role: Access/Preservation/Supplemental/Auxiliary), instead of relying on unordered `Repo.all()` results or the order file set IDs happened to be passed in.
- Added tests verifying order is preserved regardless of the order of the `fileset_ids` argument, and that a manually reordered set of file sets keeps that order when transferred into a newly created work.

- [x] Patch

# Steps to Test

Reorder a work's Access (or Auxiliary) file sets via the Structure tab drag-and-drop, then use "Transfer File Sets" on the Preservation tab to move a subset of them to a new work (and separately, to an existing work). Confirm the destination work's Structure tab shows the file sets in the same order they had in the source work.

# :rocket: Deployment Notes

- no special deployment steps needed

### DIFF
--- a/app/lib/meadow/data/works/transfer_file_sets.ex
+++ b/app/lib/meadow/data/works/transfer_file_sets.ex
@@ -384,10 +384,22 @@ defmodule Meadow.Data.Works.TransferFileSets do
 
     with {:ok, :valid} <- validate_filesets_exist(filesets, fileset_ids),
          {:ok, transferred_ids} <-
-           perform_fileset_transfer(filesets, target_work_id, max_rank_in_target_work) do
+           filesets
+           |> order_filesets_by_rank(fileset_ids)
+           |> perform_fileset_transfer(target_work_id, max_rank_in_target_work) do
       Logger.info("Transferred #{Enum.count(transferred_ids)} file sets to work #{target_work_id}")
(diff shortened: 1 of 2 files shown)


bmquinn:
I like this change! One small implementation question: could we reuse `Data.ranked_file_sets_for_work/1` here instead of duplicating the role/rank ordering logic? That helper already defines the canonical work order as `A, P, S, X`, with each role ordered by `rank`. We could fetch the source work’s ranked file sets and filter to the selected IDs before assigning the new target ranks.

My thinking is that would also make it clearer that we are preserving the source work’s existing order, and it keeps the ordering rule in one place.

bmquinn (approved):
Looks great, thanks for making that change 👍

PR #5603, opened 2026-07-20
Update to new DCAPI canvas id format

kdid:
# Summary

The "Set poster image for …" button was disappearing on VIDEO work pages shortly after the Clover viewer loaded. Root cause: the DC API's IIIF manifest canvas ids changed format (`/canvas/access/{n}` → `/file-sets/{fileSetId}?as=iiif`), but Meadow's canvas-change handler still assumed the old positional format and used the parsed value as an array index. When it failed to resolve, it wiped out the active file set in context, which also broke the "active" highlight in the file set list and canvas-switch syncing on multi-canvas works. This PR updates Meadow to resolve the active file set by matching the file set UUID embedded in the new canvas id, and hardens the reducer so a failed lookup can no longer wipe existing state.

fixes https://github.com/nulib/repodev_planning_and_docs/issues/5831

# Specific Changes in this PR
- `Viewer.jsx`: `handleCanvasIdCallback` now parses the file set UUID out of the canvas id (`.../file-sets/{fileSetId}?as=iiif`) and resolves it via `fileSets.find(fs => fs.id === fileSetId)`, instead of indexing into `fileSets` by a positional segment. Bails out (leaving the current active file set in place) if the id doesn't resolve.
- `work-context.jsx`: `updateActiveMediaFileSet` reducer now sets `activeMediaFileSet` to `null` when no file set is provided, instead of spreading `undefined` into `{}` (a truthy object with no `id` that silently broke every consumer's guard checks).
- `PosterSelector.jsx`: defensive optional chaining on `activeMediaFileSet?.coreMetadata?.label`.
- Fixed a test-mock bug across 4 test files where the Clover mock invoked a `canvasCallback` prop that doesn't exist on the real component (the real prop is `canvasIdCallback`) — this is why the broken code path had no test coverage. Mocks now also use the current DC API canvas-id format.
- Added regression tests in `Viewer.test.js` covering: the poster button survives an unresolvable canvas id, and the active file set/poster button correctly updates when the canvas id resolves to a different file set.

- [x] Patch

# Steps to Test

open any VIDEO work and confirm the "Set poster image for …" button stays visible after the viewer finishes loading (previously it be seen while the viewer is loading and then disappear). For a multi-canvas video work, confirm switching canvases in the viewer keeps the poster button and the file-set list's "active" highlight in sync with the displayed canvas.

# :rocket: Deployment Notes

- no special deployment tasks

### DIFF
--- a/app/assets/js/components/UI/IIIF/PosterSelector.jsx
+++ b/app/assets/js/components/UI/IIIF/PosterSelector.jsx
@@ -59,7 +59,7 @@ function IIIFViewerPosterSelector() {
     return null;
   }
 
-  const label = workState.activeMediaFileSet.coreMetadata.label;
+  const label = workState.activeMediaFileSet?.coreMetadata?.label;
 
   return (
     <div
--- a/app/assets/js/components/UI/IIIF/Viewer.jsx
+++ b/app/assets/js/components/UI/IIIF/Viewer.jsx
@@ -39,16 +39,18 @@ const IIIFViewer = ({ fileSets, iiifContent, workTypeId }) => {
    * When the Canvas changed in Clover, update the active media file set in Context.
    */
   const handleCanvasIdCallback = (canvasId) => {
-    if (canvasId) {
-      const canvasIndex = canvasId.split("/").pop();
-      if (fileSets[canvasIndex]?.id === activeMediaFileSet?.id) return;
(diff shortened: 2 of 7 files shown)


bmquinn (approved):
Looks good in my dev env 👍 :
Image

